PR #4, opened 2020-05-10
Added script for allen atlas generation

vigji:
Looking at BrainRender code, I made script trying to keep the essential to generate our minimal Atlas folder.

Right now it downloads the reference and the annotations  as well as all the available meshes; it saves the metadata json and a json with the annotations. I liked the way the Allen stores the region list, so right now the areas are stored in the same way. @FedeClaudi and @adamltyson, do you think this would be good as a base to interact with your software?

Notes:
 - There is no "hemispheres" stack right now, as that should be generated if required in case of symmetric atlases, and we assume having moved all data to the same orientation.
 - I think we can safely use a single entry for the resolution and size of the stack, as all raster data should match in principle
- It a pity that the nice Allen classes to handle the regions hierarchies are not in a separate package, but maybe we could fork the allensdk and have that?

### DIFF
--- a/atlases_generation_scripts/allenbrain_atlas.py
+++ b/atlases_generation_scripts/allenbrain_atlas.py
@@ -1,0 +1,83 @@
+from allensdk.api.queries.ontologies_api import OntologiesApi
+from allensdk.api.queries.reference_space_api import ReferenceSpaceApi
+from allensdk.core.reference_space_cache import ReferenceSpaceCache
+import numpy as np
+
+from requests import exceptions
+from pathlib import Path
+import tempfile
+import json
+
+import tifffile
+import pandas as pd
+
+RES_UM = 25
+
+# Temporary folder for nrrd files download:
+temp_path = Path(tempfile.mkdtemp()) / "allen_dwnl"
+
+# Generated atlas path:
+bg_root_dir = Path.home() / ".brainglobe"
+bg_root_dir.mkdir(exist_ok=True)
+
+save_dir = bg_root_dir / f"allenbrain_atlas_{RES_UM}um"
+save_dir.mkdir(exist_ok=True)
+
+# Download annotated and template volume:
+#########################################
+spacecache = ReferenceSpaceCache(
+    manifest=temp_path / "manifest.json",
+    # downloaded files are stored relative to here
+    resolution=RES_UM,
+    reference_space_key="annotation/ccf_2017"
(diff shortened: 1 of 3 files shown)


FedeClaudi:
nice one! 

For region hierarchies, I think we should have own code shared across atlases. 
We'd need to build a json/csv file with hierarchy metadata and deal with that accordingly, but it should be worth it. 

How long does it take to download all meshes? How much space to they take?

vigji:
> For region hierarchies, I think we should have own code shared across atlases.

I agree, I was just wondering if we could take as a base the allen class (without having the full package as depencency) and use a simplified version of it.

> We'd need to build a json/csv file with hierarchy metadata and deal with that accordingly, but it should be worth it.

Currently, it's a json file with a list of regions; each region has (I just removed couple of fields from Allen file):
- id: number matching the volume;
- acronym: structure short name;
- graph_order: useful if we want to allow more or less coarse gaining - it's simply n of parents;
- structure_id_path: list with parents chain
- rgb_triplet: this could be left empty but I think it's useful having the possibility of predefine it

I would actually keep this structure, and make sure we can build it for the other atlases as well - once we have the region hierarchy should be simple.

> How long does it take to download all meshes? How much space to they take?

15 mins on my wifi and they are around 200 mb. I would definitively go for having all of them stored locally together with the other stacks

adamltyson:
Cool, is this a kind of utility script to allow us to download the files we need to make our own atlas "packages"? 

I ask because I'm keen to not have this as the way that data is fetched for the user because:

- The allensdk has some annoying depedencies (pinned versions etc).
- amap requires processed template files (and some of resolutions not released by the Allen).

FedeClaudi:
I think the idea is that we will have to have some code to manually build each atlas we host, and this is it for Allen.

Once we have the data/metadata organised and hosted somewhere we will have our atlas Api fetch it from there.

We can have the allensdk   as a 'dev' requirement

adamltyson:
Perfect.

vigji:
Yes, definitively! It's temporarily there, we could even make a separate private repo with all such scripts for the various atlases!
Atlases data will be fetched only from our pre-made compressed folders, or generated locally by the users.

FedeClaudi:
Should we merge this?

adamltyson:
Yep. This is excellent (only just had a look at it properly). Merging now.

PR #11, opened 2020-05-27
Update project structure

adamltyson:
* Update setup.py with new name etc.
* Add requirements
* Format with black
* Fix failing tests
* Add testing across operating systems and deployment to PyPI with travis.


FedeClaudi:
looks good.

.travis.yml
```diff
@@ -1,41 +1,98 @@
-language: python
-matrix:
-  include:
-  - os: linux
-    python: 3.6
-    dist: bionic
-  - os: linux
-    python: 3.7
```
vigji, line 8: We don't test for python 3.8? Not urgent now, just asking in general

adamltyson:
@vigji I've made these changes without a full understanding of what you and Fede have done so far. Is it ok to merge?

setup.py
```diff
@@ -1,11 +1,11 @@
 from setuptools import setup, find_namespace_packages
 
-requirements = []
+requirements = ["allensdk", "tqdm", "numpy", "tifffile"]
```
vigji, line 3: Ugh, sorry for the allensdk requirement, I will get it out in the next round of improvements

tests/test_atlas.py
```diff
@@ -11,58 +11,66 @@ def atlas():
...
-
-@pytest.mark.parametrize("attribute, val", [("shape", [132, 80, 114]),
-                                            ("resolution", [100, 100, 100]),
-                                            ("name", "test"),
-                                            ("symmetric", True)])
+    assert atlas.metadata == {
+        "name": "test",
+        "citation": "Wang et al 2020, https://doi.org/10.1016/j.cell.2020.04.007",
+        "atlas_link": "www.brain-map.org.com",
+        "symmetric": True,
```
vigji, line 18: Yup, this is why I would like to have the atlas generation scripts within the same repo, so that we can test those consistencies together. I will move them back in my next PR

vigji:
With my next PR I promise some more documentation!

.travis.yml
adamltyson: I didn't generally test py38 because not everything supports it yet, but everything here should. I'll add it.

adamltyson: @vigji I'm having trouble installing on py38 (on ubuntu), the issue seems to be with statsmodels/numpy (even though both have py38 wheels). Have you been using this on python 3.8?

adamltyson: [Doesn't seem to be Ubuntu specific](https://travis-ci.com/github/brainglobe/brainatlas-api/builds/168447654), so I'm dropping this for now. 

I think it'll be fine when the allensdk dependency is removed.

PR #14, opened 2020-05-28
Refactoring structures/..

FedeClaudi:
Changes to `structures/simple_tree.py` and `structures/structure_tree.py`:

1. removed obsolete methods to remove dependency on allensdk
2. made most methods that expected and return a list more flexible, now you can pass a single item and they will return a single item
3. move `print_structures` and `print_structures_tree` from `core.Atlas` to `structures_structures_tree.py` to keep `Atlas` cleaner.

### DIFF
--- a/brainatlas_api/structures/structure_tree.py
+++ b/brainatlas_api/structures/structure_tree.py
@@ -41,6 +41,7 @@ import functools
 
 import numpy as np
 import pandas as pd
+from treelib import Tree
 
 from .simple_tree import SimpleTree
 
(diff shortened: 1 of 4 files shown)


brainatlas_api/structures/structure_tree.py
```diff
@@ -389,3 +396,93 @@ def path_to_list(path):
             return list(path)
 
         return [int(stid) for stid in path.split("/") if stid != ""]
+
+    def print_structures(self, to_file=False, save_filepath=None):
```
vigji, line 400: I think this is a convenient function and I see the point of having it here; I am a splitting maniac and I would keep it as a separate function then used here, put it's a matter of taste - as long as this class does not get too long

FedeClaudi: > I am a splitting maniac 

You really are :P

I prefer keeping the number of files / classes low at the cost of slightly longer classes. Especially for stuff like this which is behind the scenes and we won't have to edit it often if at all. 

I guess we can find a compromise? :)

vigji: I would agree on a maximum number of lines per class 😄  500?

PR #19, opened 2020-05-28
Brought togehter the latest changes in script generation 2

FedeClaudi:
Things left to do:

- Using @vigji 's latest code to make sure that ratatlas.py conforms to the structure and standards of the other atlas generating scrips. (@FedeClaudi )
- Making sure that mouseatlas.py and fishatlas.py use the metadata_utils.py functions to create the .txt and .csv files (@vigji )
- Adding the cration of structures.csv to metadata_utils.py (@adamltyson )


FedeClaudi:
https://github.com/brainglobe/brainatlas-api/pull/17#issuecomment-635265919

@adamltyson  I just moved it back! 😄 
What do yous prefer?

adamltyson:
> What do yous prefer?

Maybe leave it as it is for now, and we can remove anything unnecessary later. There might be a situation where someone wants to use something in here to generate their own atlas (e.g an existing atlas but at a custom resolution).

FedeClaudi:
ookay. 

It would still be on github in that case, but I see your point.

FedeClaudi:
also I don't know why the checks are failing. I haven't looked into all the tests that you've set up.

adamltyson:
They're failing firstly because it's not black formatted. Also, the flake8 thing is **very** strict, so it might be worth installing that sooner rather than later.

adamltyson:
I think before pushing, we should all make sure that:
* Black has been run
* flake8 has been run
* All tests pass locally

FedeClaudi:
>Black has been run
flake8 has been run

Any way to automate this? I keep forgetting the commands and stuff

adamltyson:
> Any way to automate this? I keep forgetting the commands and stuff

See https://github.com/brainglobe/brainatlas-api/pull/15#issuecomment-635244518

If you install with `pip install -e .[dev]`, `pre-commit` will be installed, so then you can run `pre-commit install`, and it will check the formatting with every commit (the first one will be slow as it all gets set up).

FedeClaudi:
Thanks, I'll do this and update the PR. 

This is some good stuff, I should implement it in brainrender as well!

FedeClaudi:
image

I should've fix things

brainatlas_api/atlas_gen/atlas_scripts/ratatlas.py
```diff
@@ -0,0 +1,364 @@
...
+        continue
+
+    write(extract_volume_surface(Volume(vol)), savepath)
+
+
+# ---------------------------------------------------------------------------- #
+#                            FINAL METADATA AND SAVE                           #
+# ---------------------------------------------------------------------------- #
+
+metadata_dict = {
```
vigji, line 341: The idea of the wrap-up function that I was using in the test_atlas script is to offload this metadata generation to a place where there's also a set of checks on overall structure, types, etc.

brainatlas_api/atlas_gen/atlas_scripts/ratatlas.py
```diff
@@ -0,0 +1,364 @@
...
+    "name": ATLAS_NAME,
+    "species": "Rattus Norvegicus",
+    "citation": "Swanson 2018, https://pubmed.ncbi.nlm.nih.gov/29277900/",
```
vigji, line 344: Also, I would have DOIs in the citation

brainatlas_api/atlas_gen/atlas_scripts/ratatlas.py
```diff
@@ -0,0 +1,364 @@
...
+
+    if not np.isin(np.float(region.minor), volume_data):
+        # print(f'{region.abbreviation} doesnt seem to appear in annotated dataset')
+        continue
+
+    vol[volume_data == np.float32(region.minor)] = 1
+    if np.max(vol) < 1:
+        raise ValueError
+
+    write(extract_volume_surface(Volume(vol)), savepath)
```
vigji, line 263: We should use the write_reference and write_annotation functions so that all stacks are saved in the same way

vigji:
Sorry, the comments might have been premature!

brainatlas_api/atlas_gen/atlas_scripts/ratatlas.py
FedeClaudi: yeah I need to integrate your changes still, I just copied over what I had in the other repo

brainatlas_api/atlas_gen/atlas_scripts/ratatlas.py
FedeClaudi: yep

FedeClaudi:
Do we want to add something like [interrogate](https://github.com/econchick/interrogate) to check for docstrings coverage?

PR #22, opened 2020-05-28
Unified ARA & Franklin-Paxinos atlas

adamltyson:
Still need to generate the non-terminal mesh files.

### DIFF
--- a/brainatlas_api/atlas_gen/atlas_scripts/kim_unified_atlas.py
+++ b/brainatlas_api/atlas_gen/atlas_scripts/kim_unified_atlas.py
@@ -1,17 +1,21 @@
-from allensdk.core.reference_space_cache import ReferenceSpaceCache
-
-from pathlib import Path
 import tempfile
 import json
 import tarfile
-
 import tifffile
 
-from brainio.brainio import load_any
+import pandas as pd
 
-annotations_image = Path(
-    "/media/adam/Storage/cellfinder/data/paxinos_allen/annotations_coronal.tif"
+from pathlib import Path
+from brainio.brainio import load_any
+from allensdk.core.reference_space_cache import ReferenceSpaceCache
+
+
+paxinos_allen_directory = Path(
+    "/media/adam/Storage/cellfinder/data/paxinos_allen/"
 )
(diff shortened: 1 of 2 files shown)


adamltyson:
There's a lot of code copied from the existing atlas generation scripts, so this should probably be extracted into functions.

Another thing I noticed, we're not consistent with the mesh names. The ARA and zfish meshes are `id.obj`, but the others are `abbreviation.obj`. I assume that the ID is default for the ARA, but the abbreviation is possibly more useful as we can load them directly rather than by parsing the JSON.

vigji:
My concern with the abbreviations are:
 1. current fish does not provide abbreviations (no big deal, it's just the full name though)
 2. full flexibility on strings for names/annotations might generate invalid filenames maybe?

No other concerns though, if we are sure 2 is not an issue I am happy to move to name,

FedeClaudi:
For the rat and human atlas I've been using `id+'.obj'` to keep consistency with the FishAtlas, but I don't mind either way.

adamltyson:
I copied this from the rat atlas script which deals with (2) fairly well:
```python
    savepath = str(
        meshes_dir_path
        / f'{structure["id"]}.obj'.replace("/", "-").replace("\\", "-")
    )
```

FedeClaudi:
I will write some (probably vtkplotter based) code to extract meshes + possibily beautify them so that stuff will be exported to functions.

adamltyson:
Just realised I've been an idiot. None of the atlases have the abbreviation (only mine), so I can move to the ID easily.

FedeClaudi:
Hehe. Do you want me to revert the PR so that you can update the code?

vigji:
> There's a lot of code copied from the existing atlas generation scripts, so this should probably be extracted into functions.

Agreed. We can have a single function that takes metadata parameters, a numpy reference stack, a numpy annotation stac, a list of meshes (or meshes files not having to load them?) and does the entire saving part.

vigji:
loosely related - do you guys think that we can assume ids to be always positive?

adamltyson:
> Hehe. Do you want me to revert the PR so that you can update the code?

Yes please, I'm still adding stuff. Sorry I should have marked it as draft.

PR #24, opened 2020-05-28
Unified ARA & Franklin-Paxinos atlas

adamltyson:
Continuing #22 

Added non-terminal mesh generation, but need to check that this still works properly.

### DIFF
--- a/brainatlas_api/atlas_gen/atlas_scripts/kim_unified_atlas.py
+++ b/brainatlas_api/atlas_gen/atlas_scripts/kim_unified_atlas.py
@@ -20,6 +20,9 @@ from brainatlas_api.atlas_gen.volume_utils import (
     extract_volume_surface,
     load_labelled_volume,
 )
+from brainatlas_api.atlas_gen.structure_json_to_csv import (
+    convert_structure_json_to_csv,
+)
 
 paxinos_allen_directory = Path(
     "/media/adam/Storage/cellfinder/data/paxinos_allen/"
@@ -118,6 +121,8 @@ structures.append(root)
 with open(uncompr_atlas_path / "structures.json", "w") as f:
     json.dump(structures, f)
 
+# save as csv
+convert_structure_json_to_csv(uncompr_atlas_path / "structures.json")
 
 # Create meshes
 # ######################################
--- a/brainatlas_api/atlas_gen/atlas_scripts/kim_unified_atlas.py
+++ b/brainatlas_api/atlas_gen/atlas_scripts/kim_unified_atlas.py
@@ -130,6 +130,8 @@ volume = load_labelled_volume(annotations_array)
 root = extract_volume_surface(volume)
 
 write(root, str(meshes_dir_path / "997.obj"))
+del volume
+del root
 
 non_minor_regions = []
 # First create a mesh for every minor region
--- a/brainatlas_api/atlas_gen/atlas_scripts/kim_unified_atlas.py
+++ b/brainatlas_api/atlas_gen/atlas_scripts/kim_unified_atlas.py
@@ -128,14 +128,15 @@ volume = load_labelled_volume(annotations_array)
 
 root = extract_volume_surface(volume)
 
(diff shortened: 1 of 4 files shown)


adamltyson:
As this atlas can be the same resolution as the ARA, it can be much bigger (100x) than other atlases that need the meshes extracting automatically (like the rat atlas). The mesh extraction for high resolutions can therefore be slow. Theoretically we shouldn't have to do this very often, so do we live with it taking a long time (potentially hours), or do we:
* Try to parallelise across CPU cores
* Find a quicker way to extract the surface
* Use lower resolution meshes for all atlas resolutions
* Some combination of the above

FedeClaudi:
The step that is slow, in the code I'm setting up for extracting the mesh, is the smoothing. 
Everything else happens in <1 minute.

There might be clever ways to either speedup the smoothing or to find a way to smooth the decimated mesh. Though this last step introduces more artifacts. 

Performing in parallel should also be fine, the smoothing is a bit memory hungry but it should be okay. I don't know if it's worth setting it up to run on the HPC even. 

But otherwise I don't mind it being slow. As you said it should be a one off per atlas.

adamltyson:
>The step that is slow, in the code I'm setting up for extracting the mesh, is the smoothing.
Everything else happens in <1 minute.

Is 1 minute for all regions or for one? Extracting 1400 regions for the 10um mouse atlas looks like it will take ages.

FedeClaudi:
have a look at the code here: https://github.com/brainglobe/brainatlas-api/pull/27
maybe you can find a way to speed things up

adamltyson:
This is ready to review, it all seems to work ok, but I haven't fully tested it at 10um. It should be easy to speed up via parallelisation across cores, but I need to see how much RAM it uses.

FedeClaudi:
Looks good. How do the meshes look? 

I'm testing my mesh creation code, without the smoothing the 'root' mesh takes 40s. I think that smaller meshes will be faster though.

adamltyson:
> Looks good. How do the meshes look?

They're ok, but not perfect. The original atlas was only defined at 50um in Z. When Steve and I made the 10um atlas, we didn't interpolate in Z, so it's a bit blocky in that axis.

Screenshot from 2020-05-29 13-24-57

adamltyson:
I think as it is currently, the mesh extraction uses too much RAM to be usefully parallelised.

FedeClaudi:
The blockiness might be improved using `mesh_utils.extract_mesh_from_mask` with `smooth=True` from https://github.com/brainglobe/brainatlas-api/pull/27

adamltyson:
I've added a function to convert the json to a cellfinder-compatible csv. I haven't added this to the other atlas generation scripts yet, assuming we're going to extract common functionality from these.

FedeClaudi:
nice, looks good. We can add to `atlas_gen.metadata_utils` a function to save the .csv file with the other text files. 

Can the structures in the .csv file be in alphabetical order?

adamltyson:
> Can the structures in the .csv file be in alphabetical order?

Done

FedeClaudi:
Should we merge? 
Also, are you working on code to make meshes for non-terminal nodes or do you want me to?

adamltyson:
Yep, this is just finished. This code generates meshes for terminal nodes, it can probably be copied into the other atlases, but I haven't fully checked that everything works.

FedeClaudi:
it will need some changes because I've improved the mesh generation function, but it should be just a matter of changing the name function.  I'm also adding methods to look at meshes and volume side by side to make sure it looks good.

adamltyson:
The acronyms in this atlas are a bit weird, as is the hierarchy. Also considering the meshes are probably not going to be as nice as the ARA ones, somewhere we should suggest that people use the ARA unless they specifically need the extra segmentation.

adamltyson:
Are you happy if we merge this, then we merge your PR?

FedeClaudi:
yeah let's merge this first, mine needs some more work

PR #29, opened 2020-05-31
Moved atlas_gen to separate package

vigji:
In this PR:
- I moved the atlas_gen module to be a separate package; this in principle should make sure there's no confusion on what should be imported in the main package that needs to be tested;
- I removed 2 of the tests that could not be parallelelized;
- I removed the installation of the dev tools for the tests.


vigji:
Final test structure: 
 - Linux (one version only): py3.6, 3.7, 3.8;
 - Windows: py3.7 only;
 - Separate test for linting, so black/flake8 errors are lumped together in a single place.

Tests are significantly faster now, although Windows still take some more time to setup. Maybe we could temporarily disable that together with macOS? I left all the code in the travis folder though, so put them back in the future will be straightforward.

I would merge this now to work on some other things but @adamltyson let me know if this is fine for you!

adamltyson:
I won't have time to review this properly until tomorrow, but it looks ok. 

I'm keen to keep as much testing as we can though, because I've come across python version/OS interactions in the past, when wheels are only built for specific versions etc.

>  although Windows still take some more time to setup. Maybe we could temporarily disable that together with macOS? 

I think we should absolutely keep the Windows test, and I would like to bring the macOS test back if we can. 

Ideally we would also test every combination, but as this takes a long time, maybe we can keep it for tagged commits (i.e. to prevent deployment if there is an issue)?

Sorry to be a pain, but I think the tests taking a bit longer isn't much of an issue if we make sure everything passes locally (and maybe be selective with the commits we push?). I'm also wary that I will be maintaining quite a lot of software that will likely depend on this package, so I'm keen to test as much as is practical.

vigji:
Ok, I added back the macOS test.

True that testing locally should take care of most problems. Still, maybe all the matrix is an overkill for all commits. If you know a way of trigger the job will all OSs and all versions only on marked commits that would be great!

vigji:
I will marge for now as I don't see potential issues, but let me know if something's not ok. I am happy to reintroduce the grid testing if it can restricted to tags in the future!

adamltyson:
Cool, I've added this functionality in #30.

PR #30, opened 2020-06-01
Refactor tests

adamltyson:
This PR:
* Removes a dependency specified twice (brainio).
* Only reports to coveralls on one test step, speeding up the tests (slightly) and 
prevents the linting step from submitting to coveralls (and incorrectly reporting 0% test coverage).
* Adds additional tests (macOS and Windows py36 & py38) that only run on tagged commits. So normally, only the 5 tests run, but when we want to make a release, the order for a tagged commit will be:
    * Normal tests
    * Additional tests for Windows and macOS to ensure no weird OS/Python version interaction. These tests will only run if the "normal" tests pass, so we can fail quicker. 
    * Deploy to PyPI, if and only if the first two steps pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,13 @@ after_success:
 
 stages:
   - test
+  - name: deploy_test
+    if:
+      tag IS present
   - name: deploy
     if:
       tag IS present AND repo = brainglobe/brainatlas-api
+
 
 
 jobs:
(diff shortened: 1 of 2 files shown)


adamltyson:
I think this ok for now, as I think all the "normal" tests are running quickly in parallel. If we come across issues, we could do stuff like only testing across OSs on PRs etc.

vigji:
Perfect, looks great to me!

adamltyson:
Merging now, as it all looks fine. Worth mentioning that we can any additional pre-deployment tests.

PR #31, opened 2020-06-01
Centralising all atlas saving

vigji:
I realised that having some checks on the files and save in a fixed ways only the "light" files - metadata and structures - does not make that much sense, as I think it's a safe assumption that all of us can fit in memory reference+annotation+meshes and pass all of them as variable to a function.

This is a little inefficient in I/O (eg, you download meshes, read all of them, then pass them again for saving); but efficiency should not be a big deal within the occasional atlas generation, and having a single function that does all the saving has huge advantages for making sure that every atlas has truly consistent entries. Moreover, makes Atlas versioning way easier: mayor version is associated with a particular version of the saving function+its calls, and it is univoke for all atlases generated within the same code version. Finally, in the long term, it will be easy to maintain those atlases and update them to new versions: we can just load the old atlas data, and use the saving function with whatever correction to save the new version.

I am opening this as a draft now, hope to be done as soon as possible. but this way you can have a look and in case discuss more stuff.

Current points to be considered:
- meshes: I am thinking of passing them as a dictionary of {meshfilename: (vertices, edges (faces?))} and use something to save the obj;
- orientation: I am lining toward passing to the function the atlas orientation defined as a valid origin for [BGSpace](https://github.com/brainglobe/bgspace), and handle in the function stack+mesh reorientation; this can easily be overwritten if one want to handle the orientation in the atlas script simply by doing manually the reorienting and then passing an orientation that matches the standard one (CCF) found in the descriptors.


adamltyson:
>  as I think it's a safe assumption that all of us can fit in memory reference+annotation+meshes and pass all of them as variable to a function.

The `kim_unified` atlas might push this, it's >20GB.  

Is the idea that you would download everything, check and resave? Couldn't this take a very long time over wifi when we're at 100GB+ of atlases?

FedeClaudi:
But this is just for atlas generation right? I don't see why it would take any longer then normal, you still have to download/generate all data before saving.

adamltyson:
> eg, you download meshes

Maybe I'm confused. I saw this, and assumed we were downloading the meshes that we'd generated previously.

FedeClaudi:
I'm confused as well haha :P

vigji:
Sorry for the confusion - yes, this is only for atlas generation and it won't be any more downloading, just more reading/writing.

Currently the workflow in my scripts for generating the atlas is:
 -> fetch raw data remotely (stacks, meshes);
 -> fix only what need to be fixed (eg, reorient the stacks, but leave the meshes in one place if they are already readable .obj);
-> finalise the folder saving metadata.

I would move this to:
-> download everything & load (in RAM);
-> pass everything to a single saving function for saving.

If loading everything in ram is a problem (only for the stacks I guess?) we can have the option of running the stack loading+reorientation+saving in a sequential way (not passing a np.array but a tiff file that can be loaded).

adamltyson:
> If loading everything in ram is a problem (only for the stacks I guess?) 

Not necessarily, It can probably be improved with smoothing etc, but the meshes for the kim atlas are much bigger than the stacks actually (>10GB).

This sounds good though. We should presumably have somewhere to store the "raw" data, i.e. the data that we use to generate the atlas (if it's not already online)?

FedeClaudi:
I see, what's the advantage of loading everything at once?
If you save all meshes as  .obj in a folder, why not passing the folder and loading them as needed?

vigji:
> This sounds good though. We should presumably have somewhere to store the "raw" data, i.e. the data that we use to generate the atlas (if it's not already online)?

I agree, maybe each bundle of raw data could go to a dedicated GIN repo. I also hope that once we have a first stable version of all atlases this won't be a problem anymore as we can generate new versions just building on the previous ones - obviously under the assumption that we don't need new data.

vigji:
> I see, what's the advantage of loading everything at once?
> If you save all meshes as .obj in a folder, why not passing the folder and loading them as needed?

That you pass the meshes in a single consistent format, and you deal with the variegation of mesh specifications format in your script instead of dealing with multiple scenarios in this function

vigji:
Sorry - yes, if you saved them as .obj in your script that would provide consistent files - but that would in turn be inefficient if we want to save things again in this function. What I am suggesting is:

mesh generation/loading -> all meshes in RAM in a dictionary -> all meshes written in centralised function as .obj

what I think you are saying is:

mesh generation/loading -> all meshes saved as .obj in script -> all meshes read and written in centralised function as .obj

It would work but save an unnecessary intermediate .obj file  - in the paradigm of this PR, where the aim is to have all the final files saved with the same lines of code.

FedeClaudi:
I see, okay. 
The reason why I like saving an intermediate .obj file is that when you are generating the meshes from images (like in the human atlas) this can take a while. So when you're debugging/developing other parts of the code it helps to save the meshes and avoid generating them anew. But I guess this can be done anyways and then everything is loaded before saving. 

----------
Btw unrelated but I think we should use  [meshio](https://github.com/nschloe/meshio) to load/save .obj files. vtkplotter is useful for creating and editing meshes, but for simple I/O we can use meshio. It also has a function to compress mesh files which me might try to reduce file size. 
I will also test [pygalmesh](https://github.com/nschloe/pygalmesh) as an alternative to vtkplotter for creating meshes from images.

vigji:
ok, great, meshio looks great!

Actually, if we use it, you guys are right: we don't have to load all files, just pass a dictionary of mesh id / meshio compatible file path; load them with default meshio function and automatic -type inference, and then save from meshio as obj files. I guess this should ensure enough consistency?

adamltyson:
> Actually, if we use it, you guys are right: we don't have to load all files, just pass a dictionary of mesh id / meshio compatible file path; load them with default meshio function and automatic -type inference, and then save from meshio as obj files. I guess this should ensure enough consistency?

I think this is good. 

More generally, I think we should avoid loading things into RAM if we don't need to. Its possible that the Allen CCF V4 could be defined in a 1um space, with more regions, and the resulting files become 100GB+. Unlikely, I know, but we shouldn't make it impossible to use.

FedeClaudi:
>Actually, if we use it, you guys are right: we don't have to load all files, just pass a dictionary of mesh id / meshio compatible file path; load them with default meshio function and automatic -type inference, and then save from meshio as obj files. I guess this should ensure enough consistency?

sounds like a good plan

vigji:
Ok, so maybe let's support the option for the stacks to be filenames of tiff files to be loaded with tifffile if they are not numpy arrays?

FedeClaudi:
ok

vigji:
So, there's a version that's fully running right now. I tested it on the example, and will generate the Allen and the fish brain again this afternoon, but in the meantime if you want you can have a look - this is basically final.

FedeClaudi:
Thanks, I'll update humanatlas with the new code

vigji:
Everything seems to have run smoothly for fish and Allen atlas. I would merge this so that you can work with the function from the master

PR #38, opened 2020-06-14
Added config file to handle default directories

vigji:
Figuring out configurable file for the brainglobe directory. Also:
 * downloading bar when retrieving the atlas;
 * name fixes for the atlases

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-bg_config.conf
+brainatlas_api/bg_config.conf
 
 workspace.py
 .vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+bg_config.conf
+
 workspace.py
 .vscode/
 
--- a/atlas_gen/metadata_utils.py
+++ b/atlas_gen/metadata_utils.py
@@ -11,8 +11,8 @@ from brainatlas_api import descriptors
 import requests
 from requests.exceptions import MissingSchema, InvalidURL, ConnectionError
 
-from brainatlas_api.structures.structure_tree import StructureTree
 from atlas_gen.structure_json_to_csv import convert_structure_json_to_csv
+from brainatlas_api.structure_tree_util import get_structures_tree
 
 
 def generate_metadata_dict(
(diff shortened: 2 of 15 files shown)


vigji:
This should now be fixed.

The way it currently works: There is a bg_config.conf file in the repo. Its position is kept in the variable CONFIG_PATH here: https://github.com/brainglobe/brainatlas-api/blob/8f60bb9c0d41f42bf91e80c20d561ba9aa1ef865/brainatlas_api/config.py#L6

If you can suggest a better way of referencing to the package location, I am happy to change this.

There is no "template" or "dafault" config in the repo, which I can imagine being a pain for version control of we want to avoid weird file duplications. It is written from a standard configuration dictionary here:
https://github.com/brainglobe/brainatlas-api/blob/8f60bb9c0d41f42bf91e80c20d561ba9aa1ef865/brainatlas_api/config.py#L10

so that it is initialised with the valid user root folder. The folder is created in the BGAtlas class the first time we actually need it:

https://github.com/brainglobe/brainatlas-api/blob/8f60bb9c0d41f42bf91e80c20d561ba9aa1ef865/brainatlas_api/bg_atlas.py#L44

To change this location, there's 3 ways:
 - manually editing the `bg_config.conf` file;
 - using `brainatlas_api.config.write_config_value(brainglobe_dir, new_path)`;
 - using a little command line utility: 
```
 > brainatlas_config -k brainglobe_dir -v /Users/luigipetrucco/.brainglobe
```
In the future the configuration could be used to handle other values.

Let me know what you think about all of this, also eg for variable names etc, it would be important to keep this part of the code stable in the future.

FedeClaudi:
Nice one, it looks really good. 

>using a little command line utility:

What's the best way to add tests for this kind of stuff?

Currently `bg_atlas` accepts a `brainglobe_path` argument:
https://github.com/brainglobe/brainatlas-api/blob/8442627ee7b9df35958e7d8989737c956d0d9503/brainatlas_api/bg_atlas.py#L34

How does it interact with the path in the config stuff?

vigji:
> Nice one, it looks really good.
> 
> > using a little command line utility:
> 
> What's the best way to add tests for this kind of stuff?

If we agree that we like this scheme I will try to figure out tests, click has some testing modules so it should not be difficult.

> 
> Currently `bg_atlas` accepts a `brainglobe_path` argument:
> 
> https://github.com/brainglobe/brainatlas-api/blob/8442627ee7b9df35958e7d8989737c956d0d9503/brainatlas_api/bg_atlas.py#L34
> 
> How does it interact with the path in the config stuff?

Object arguments have full priority over default ones!

adamltyson:
Looks good. 

> If you can suggest a better way of referencing to the package location, I am happy to change this.

I'm not sure if it's a "better" way to reference the package location, but you can also use something like this, that would not be dependent on the location of the file calling the function:
```python
from pkg_resources import resource_filename
CONFIG_PATH =  Path(resource_filename("brainatlas_api", "bg_config.conf"))
```

vigji:
> ```python
> resource_filename
> ```

Great, I like this more! Then I'll write tests and merge this by today.

Issue #10, opened 2020-05-27
Versioning atlases


vigji:
I think we should also have a way of nicely integrating versioning of this package together with versioning that I hope GIN can provide on the packages deployed there. If we introduce new features in the API they could require the latest version of an atlas to be redownloaded for the user that has an outdated one. Viceversa, maybe sometimes we should ensure that outdated API versions ask for legacy versions of the atlases.

@adamltyson @FedeClaudi any thought on this? Does this make sense?

adamltyson:
That's a good idea, and it's an issue that I've come across with amap and cellfinder. Until now I've avoided it because they're not yet at a stable version. However, this project could reach stability, and still undergo lots of changes to the atlases.

Do we want to start trying to support old versions of the API/atlas? This could be a headache avoided with a message asking people to upgrade.

FedeClaudi:
I think for now, since things will be changing a lot and rapidly we can just ask people to us the updated version. Once development slows down we can start supporting different versions?

vigji:
Youp, I also agree that priority is to automatically detect if it is outdate and trigger redownloading. I was not thinking about supporting old versions of the atlases, just to somehow map in the `brainatlas-api` version the version of the atlases it is working with. This way matching the correct (old) version of the atlas to be used is not that different from updating to the most recent one

vigji:
In this direction, probably it make sense to put the scripts to generate the atlases back in a separate folder of this repo, so that they are versioned together?

adamltyson:
Could each atlas just have unique metadata entry, so we know what version of what atlas we are using (like `Allen_10um_2020-05-27a`)?

We can prompt people to upgrade to a new version of the API (and therefore atlases) with something like this in `__init__.py`
```python
__name__ = "brainatlas-api"

import luddite
from packaging import version

most_recent_version = luddite.get_version_pypi(__name__)

if version.parse(__version__) < version.parse(most_recent_version):
    print(
        f"This version of {__name__} ({__version__}) is not the most recent "
        f"version. Please update to v{most_recent_version} by running: "
        f"'pip install {__name__} --upgrade'"
    )
```

adamltyson:
> In this direction, probably it make sense to put the scripts to generate the atlases back in a separate folder of this repo, so that they are versioned together?

Maybe, but this could stop being useful as soon as anyone else makes an atlas that we incorporate.

FedeClaudi:
I think we are happy with how things are now? Should we close?

PR #44, opened 2020-06-27
Atlas version control - first draft

vigji:
We still lack a way of properly addressing local vs. remote atlases versions, which might be important to tackle consistently from the beginning (related to #10 ).

I made a draft of some changes to the BGAtlas class, so that:
 - there is no more `self.version` attribute defined manually in the class;
 - major version is determined from a global descriptor in the descriptors module. If at any point something changes in brainatlas-api which is no more back-compatible with older atlases, this should be increased;
 - most recent available minor version is determined from an heuristic search on the GIN urls (try urls with different versions and find the one that produces a correct page response). I truly think that something like this would be important - having reference to the most recent remote version of the atlas - but the current approach has some drawbacks:
      * it is ugly;
      * it is slow (about 250-300 ms per request, which - with a break - would mean about 1-2 sec every time we instantiate an Atlas object);
So I am happy to hear any input/other approaches. One option could potentially be refactor what now was the self.version into a `self.most_recent_version` and refer to that instead of using this search, under the assumption that we the curators always keep these numbers in good synch.  Let me know what you think on this point
- If local major version is behind minimal required major version, an error is raised; if local minor version is behind last available minor version for the atlas, a warning is raised. 

I would now add an option to pass in the init to automatically remove the old atlas and download a newer one, if available. This might also solve the speed problem I raised above, making the control happening only if a `check_updates` argument is passed in the class constructor

### DIFF
--- a/atlas_gen/wrapup.py
+++ b/atlas_gen/wrapup.py
@@ -19,7 +19,7 @@ from brainatlas_api import descriptors
 
 # This should be changed every time we make changes in the atlas
 # structure:
-ATLAS_VERSION = 0
+ATLAS_VERSION = descriptors.ATLAS_MAJOR_V
 
 
 def wrapup_atlas_from_data(
--- a/brainatlas_api/__init__.py
+++ b/brainatlas_api/__init__.py
@@ -8,7 +8,8 @@ available_atlases = [
 
 def get_atlas_class_from_name(name):
     names = [
-        f"{atlas.atlas_name}_v{atlas.version}" for atlas in available_atlases
+        f"{atlas.atlas_name}_v{atlas.local_version}"
+        for atlas in available_atlases
     ]
     atlases = {n: a for n, a in zip(names, available_atlases)}
 
--- a/brainatlas_api/__init__.py
+++ b/brainatlas_api/__init__.py
@@ -7,10 +7,8 @@ available_atlases = [
 
 
 def get_atlas_class_from_name(name):
-    names = [
-        f"{atlas.atlas_name}_v{atlas.local_version}"
-        for atlas in available_atlases
-    ]
+    names = [atlas.atlas_name for atlas in available_atlases]
+
     atlases = {n: a for n, a in zip(names, available_atlases)}
 
     if name in atlases.keys():
--- a/brainatlas_api/bg_atlas.py
+++ b/brainatlas_api/bg_atlas.py
@@ -1,12 +1,9 @@
 from pathlib import Path
 import tarfile
-import requests
-import warnings
 
 from brainatlas_api import utils
 from brainatlas_api import config
 from brainatlas_api import core
-from brainatlas_api import descriptors
 
 
 COMPRESSED_FILENAME = "atlas.tar.gz"
@@ -19,6 +16,10 @@ __all__ = [
     "KimUnified25Um",
     "KimUnified50Um",
 ]
+
+
+def _version_tuple_from_str(version_str):
+    return tuple([int(n) for n in version_str.split(".")])
 
 
 class BrainGlobeAtlas(core.Atlas):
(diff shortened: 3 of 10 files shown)


FedeClaudi:
what about having a .json file on GIN with just the version of the latest atlas?
That way you only need a query. We'd have to manually keep it updataed but it would make things easier?

Or we could have it on github even.

vigji:
If we want to go with manual updating of a file, we can put it back as an attribute of each class and that would be the most straightforward way!

FedeClaudi:
Related to the topic: we should have some way to let users 'update' their atlases. Currently if a users doesn't have local files for an atlas the atlas is automatically downloaded. I think we should have something that checks if a more recent version of the atlas exists and lets the user know so that they can update it. 

Do we already have a command line command for downloading/updating atlases? If not, we should. Then when we detect that the user has an outdated version of the atlas we can suggest they use the command to update it.

vigji:
Yesssnj, I was thinking of implementing something like that in this PR!
I was thinking mostly of giving arguments to the bgatlas class for this, but you are right, line commands make perfect sense here!

FedeClaudi:
I'm happy to do the update stuff, but we need to finilise how we handle versions. 

> If we want to go with manual updating of a file, we can put it back as an attribute of each class and that would be the most straightforward way!

The reason why I thought having a metadata file  is best, is that the version specified in the class definition it depends on the code's version. This means:
- we'd have to release a new version of the package when we change the resolution, though the code itself would not have changed necessarily
- users that don't want to update the their version for whatever reason can't get the newst atlas versions. Users that want the latest code can't use older atlas versions
- if an user updates their code, but not the atlas there would be a mismatch between the stated version and the actual one. 

The best way around this is to separate metadata and code, as we did for all the rest. 
What we could do is:
- add a `version` entry to the metadata when we generate the atlas. This means that you always know which atlas version  you have downloaded regardless of which version of the api package you're using. 
- add a `latest versions` file on dropbox or gin where we keep note of what is the latest version of each atlas. This is just for convenience as it's then very easy to download this and compare to the users' version. 


Makes sense?

vigji:
> I'm happy to do the update stuff, but we need to finilise how we handle versions.

Ok thanks, then let's settle this here, I will update it and fix the tests.

> > If we want to go with manual updating of a file, we can put it back as an attribute of each class and that would be the most straightforward way!

> * we'd have to release a new version of the package when we change the resolution, though the code itself would not have changed necessarily

I think this is a very valid point. 
Although keep following this very valid point, we probably should also drop the classes MouseAtlas, etc., and replace them with a single BGAtlas class with atlas name as argument? How would people feel about that?


> The best way around this is to separate metadata and code, as we did for all the rest.
> What we could do is:
> 
> * add a `version` entry to the metadata when we generate the atlas. This means that you always know which atlas version  you have downloaded regardless of which version of the api package you're using.

This we already have in the metadata json
> * add a `latest versions` file on dropbox or gin where we keep note of what is the latest version of each atlas. This is just for convenience as it's then very easy to download this and compare to the users' version.

This is definitively a good way, a bit annoying to do by hand but I guess we can manage :)

FedeClaudi:
> Although keep following this very valid point, we probably should also drop the classes MouseAtlas, etc., and replace them with a single BGAtlas class with atlas name as argument? How would people feel about that?

Other than the fact that I use the calsses in `bgatlas` for the list atlases stuff I don't see why not.  I guess if we have a file with atlas -> version we can just use that. 

> This we already have in the metadata json

I had no idea! Then why do we have it in the class too?

vigji:
> I had no idea! Then why do we have it in the class too?

Because before this discussion that was the point of the you would have updated to make sure that a different version was downloaded, as the metadata are accessible only once you get the atlas package.

@adamltyson  would you be in favour of dropping the atlas classes in favour of a single one + constructor argument?

FedeClaudi:
Is this good to merge? For writing the update stuff...

adamltyson:
> @adamltyson would you be in favour of dropping the atlas classes in favour of a single one + constructor argument?

This is fine by me, and I think it's all good to merge.

vigji:
Let's merge it then and do that change separately, as I can't work on this again till this evening

PR #49, opened 2020-06-30
Fede dev

FedeClaudi:
I've made a few changes..
-----
When downloading files the progress bar is now implemented with rich [on windows the looks are not the best, but on linux and macOs it should be better]:
image

I've removed `tqdm` from the requirements, it's only used in the atlas generation stuff now. 
----
**atlases list** now doesn't show the remote path anymore, but I've added the local and online version of the atlas. 
image

Also if an an atlas is outdated this is highlighted in the table (shown for an example here):
image

----
**atlas update**
When instatiating at atlas class it checks if it's the latest version, if not it prompts the user to update

image

Using the update command results in the current atlas folder being deleted and the new atlas being downloaded:
image

if `force` is not used, the atlas is not updated if the version matches the latest version online.


adamltyson:
Hi @FedeClaudi, I ran this locally (after reinstalling, and deleting `~/.brainglobe`). I installed an atlas with `brainatlas_update -a allen_mouse_25um`, but then `brainatlas_list_atlases` gives me:

```
(brainatlas_api) adam@garfield:~/projects/brainatlas-api$ brainatlas_list_atlases
                                                                                                  
                              Brainglobe Atlases                               
╭──────────────────┬────────────┬───────────────┬────────────────┬────────────╮
│ Name             │ Downloaded │ Local version │ Latest version │ Local path │
├──────────────────┼────────────┼───────────────┼────────────────┼────────────┤
│ allen_mouse_25um │    ---     │      ---      │      0.2       │ ---        │
╰──────────────────┴────────────┴───────────────┴────────────────┴────────────╯
```

Am I doing something wrong?

FedeClaudi:
image

no, it looks like there's something off...

adamltyson:
Bingo, looks great now!

vigji:
Looks good, will test as soon as I can!
Just a note looking at the code - could we use click for out command line functions? Is it very neat to read, just add a couple of function decorators to parse the arguments

FedeClaudi:
> could we use click for out command line functions? Is it very neat to read, just add a couple of function decorators to parse the arguments

I don't understand, what do you mean with `out command line functions` and what's `click`?

vigji:
Sorry, *our* command line functions like this one for atlas update :)
Click is a library for parsing command line arguments for command line applications it’s very straightforward to use - not good for many parameters, but here we have one or two. I used it for the config module!

FedeClaudi:
I see, instead of argparse we should use click?

vigji:
Yep!

FedeClaudi:
aaaaand, done. That's cool I didn't know about it.

brainatlas_api/bg_atlas.py
```diff
@@ -59,12 +64,17 @@ def __init__(self, brainglobe_dir=None, interm_download_dir=None):
...
+            rprint(
+                f"[magenta2]Bgatlas_api: {self.atlas_name} not found locally. Downloading...[magenta2]"
+            )
             self.download_extract_file()
 
         # Instantiate after eventual download:
         super().__init__(self.brainglobe_dir / self.local_full_name)
 
+        # Compare atlas local version with latest online
+        self.check_lateset_version()
```
vigji, line 76: typo

brainatlas_api/update.py
```diff
@@ -0,0 +1,63 @@
...
+
+    # Check that everything went well
+    atlasclass()
+    rprint(
+        f"[b][magenta2]Brainatlas_api: {atlas.atlas_name} updated to version: "
+        + f"{_version_str_from_tuple(atlas.remote_version)}[/magenta2][/b]"
+    )
+
+
+@click.command()
```
vigji, line 59: These could probably work directly on the update atlas, and be used only when running it from command line

brainatlas_api/utils.py
```diff
@@ -36,19 +44,37 @@ def check_internet_connection(
 
 
 def retrieve_over_http(url, output_file_path):
+    # Make progress bar object
+
+    progress = Progress(
```
vigji, line 49: As per issue, I fear that this could not work in a notebook...

tests/test_update_atlas.py
```diff
@@ -0,0 +1,19 @@
+from brainatlas_api import update
+from click.testing import CliRunner
+
+
+def test_update():
+    update.update_atlas(atlas_name="allen_mouse_25um")
```
vigji, line 6: Maybe better to use the example as it is much faster to download

brainatlas_api/update.py
FedeClaudi: it was giving me an error when testing stuff.. not sure if Im doing something stupid with pytest.

tests/test_update_atlas.py
FedeClaudi: adding it

brainatlas_api/bg_atlas.py
FedeClaudi: fixing

brainatlas_api/update.py
vigji: Ok, then leave it this way and I'll have a look!

tests/test_update_atlas.py
vigji: Also, here it would also be important that apart from not failing running the code produce the correct results - in this case, maybe create a dummy folder with an outdated version in the name and check if it gets updated or not

FedeClaudi: it downloads the latest version with the standard `BrainGlobeAtlas` functionality, so if something like that goes wrong it wouldn't be update's fault...

brainatlas_api/update.py
FedeClaudi: I had to do the same for the show_atlas, had to make a `cli_show_atlas` otherwise the tests fail...

FedeClaudi:
among other things, `brainatlas.list_atlases.list_atlases` is not `brainatlas.list_atlases.show_atlases`

vigji:
Great! Good to merge for me.

Issue #45, opened 2020-06-29
Dealing with inputs in microns vs voxel spaces


FedeClaudi:
Currently functions like https://github.com/brainglobe/brainatlas-api/blob/bece6190d0feb3118a5b09feb4d5d93ab5499d56/brainatlas_api/core.py#L123

Accept input coordinates expressed in voxel space. We should add the possibility to pass inputs as `microns` and use the atlas' `resolution` to convert between the two. This would make life easier when using meshes (which are expressed in microns) as starting point, and it might be more intuitive for users. 

Ideally we'd let users pass the coordinates as they want, and give them an option like `as_microns`. Then behind the scenes we use the atlas' resolution to convert from microns to voxels and vice versa.

PR #51, opened 2020-07-01
Added microns option for coordinates

vigji:
Addressing #45.
- Added microns option for coordinates;
- refactored `atlas.lookup` @FedeClaudi  to `atlas.lookup_df` - we could even set it to `atlas.lookup_dataframe` for being more informative;
- listed ExampleAtlas together with the rest in __all__

### DIFF
--- a/bg_atlasapi/bg_atlas.py
+++ b/bg_atlasapi/bg_atlas.py
@@ -10,6 +10,7 @@ from bg_atlasapi import core
 COMPRESSED_FILENAME = "atlas.tar.gz"
 
 __all__ = [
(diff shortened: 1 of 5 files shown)


FedeClaudi:
awesome, did you test it with inputs both in voxels and microns? Do we have something for this in tests?

vigji:
> awesome, did you test it with inputs both in voxels and microns? Do we have something for this in tests?

Yes, testing queries on annotation and hemispheres with both units!

FedeClaudi:
you're the best!

PR #53, opened 2020-07-01
brainglobe command

FedeClaudi:
I've unified all command line commands under a single command, as suggested by @adamltyson 

Now you can just use `brainglobe` followed by
* `list` to print all atlases
* `update` with `-a atlas_name -f` to update and atlas (`-f` forces the update)
* `install` with `-a atlas_name` to install a new atlas
* `config` to modify the config.


FedeClaudi:
not sure why the commit history is so long, it should be only the last 2...

adamltyson:
> not sure why the commit history is so long, it should be only the last 2...

I think this is usually when you continue to work on a branch, that's previously been merged. It's not even with master because of the merge commit. I think you can either make a new branch, or merge master into your branch.

FedeClaudi:
I did merge master?

image

adamltyson:
> I did merge master?

No idea then, sorry.

adamltyson:
Also this is great, somehow much more professional.

> update with -a atlas_name -f to update and atlas (-f forces the update)
install with -a atlas_name to install a new atlas

What happens if you don't use the `-a` flag? Does it update  all atlases, or can we get rid of it?

FedeClaudi:
if you don't use `-a` it throws an error because you didn't pass any atlas name.

adamltyson:
Oh sorry, I meant is there any functionality without it? Or can we just do `brainglobe install atlas_name`?

FedeClaudi:
yeah `list` and `config` don't take `atlas_name` as arguments so I think to remove `-a` with `click` I would have to make it so that you always have to pass a name. Maybe @vigji knows better how to do this?

adamltyson:
Ahh i didn't know you were using click. In that case dw. It's not a priority

vigji:
I am not an expert, but maybe these groups can be helpful in this context? https://click.palletsprojects.com/en/7.x/commands/

FedeClaudi:
I only skimmed through it, but I think for now it;s good as is. Should we merge?

adamltyson:
Yep, I think this is good to merge whenever.

PR #58, opened 2020-07-13
Instantiating all Atlases from a single class

vigji:
1. As per the title
2. Some other minor refactoring and separations in list_atlases that @FedeClaudi might want to have a look to; now `get_downloaded_atlases()` return a list of actually downloaded atlases, and its old function is covered by `get_atlases_lastversions()` (happy to refactor that).
3. Improving coverage;
4. Improved code docs and docs consistency here and there;


FedeClaudi:
hey, 

That sounds good!
What's the motivation for having both `get_downloaded_atlases` and `get_atlases_lastversions`? As opposed to having only one function with the behaviour governed by a `kwarg`?

Did anything change with `show_atlases`?

vigji:
I would prefer to differentiate a purely locally working function such as downloaded atlas from somehing that requires remote fetching with internet connection

vigji:
And I will make sure that show_atlases behave the same

FedeClaudi:
makes perfect sense! It seems to be working fine :)

FedeClaudi:
Hey, 

This seems to fail just because the test `test_config_edit` expects to use the example atlas class. 
We should try to fix and merge, after that I can try working on https://github.com/brainglobe/bg-atlasapi/issues/57

PR #60, opened 2020-07-23
Refactoring atlas generation for automatic update of atlases

vigji:
* Refactoring atlas generation code in script-like functions, for easy call from a global script;
* Adding option for supplementary references (in fish there's 2-3 different transgenic ines people use for morphing); what should be the name? "secondary_references"? "supplementary_references"? "additional_references"?


adamltyson:
> what should be the name? "secondary_references"? "supplementary_references"? "additional_references"?

Don't think it matters too much, but I like "additional_references"

vigji:
good!

FedeClaudi:
Are we happy with the mesh generation code? Would it be wort looking into options like [this](https://github.com/nschloe/optimesh) to improve smoothing?

vigji:
You are the expert there :D

vigji:
@FedeClaudi I had a first superficial try of optimesh and was not really working great for my mesh, the result had broken normals and it was also taking a long time...have you already used it? It is not very documented, maybe I am just using it wrongly

FedeClaudi:
no I haven't I just saw it yesterday.. if it doesn't work it doesn't work... I guess the only way to get real nice meshes would be to import them in blender and clean them up manually...

vigji:
Indeed, that’s what I end up doing for the main one as well :)

vigji:
Test are fixed  now! If you both want to take a look, but this should be ready to merge. 
ATM tests will crash once we update the atlases, asr orientation will need to be fixed afterwards. 

The code made here has been moved to https://github.com/brainglobe/bg-atlasgen

bg_atlasapi/core.py
```diff
@@ -46,35 +48,51 @@ def __init__(self, path):
...
+            )
+
         self._annotation = None
         self._hemispheres = None
-        self._hierarchy = None
         self._lookup = None
 
     @property
     def resolution(self):
-        """Make resolution more accessible from class.
```
adamltyson, line 62: I was about to suggest we added these properties. Awesome!

setup.py
```diff
@@ -8,24 +8,6 @@
     version="0.1.1",
     description="A lightweight python module to interact with atlases for systems neuroscience",
     install_requires=requirements,
-    extras_require={
```
adamltyson, line 11: Have these been moved anywhere? They weren't just for atlas generation, but also to allow contributors to easily set up a development environment (with pytest etc.).

FedeClaudi (approved):

setup.py
vigji: Ugh right, sorry! I was just thinking about the atlas generation. I'll move those backs

vigji:
I would merge then!

PR #61, opened 2020-07-27
Fede dev

FedeClaudi:
Added a couple functions to `core` to get the ancestors and descendants of a given structure. 
Should be fairly straightforward and it passes tests locally.


FedeClaudi:
I'm again not sure why there's all of these commits and why it marks a couple files as being added... 
It might be easier if we just copy paste the functions/tests I've added directly into `master` and then I trash the `FedeDev` branch which seems to be all messed up...

vigji:
I would not mind! Just one thing: test at line 128 I assume you are testing on full allen? I don't remember such regions in the example

FedeClaudi:
aha! I thought the example atlas had metadata about all regions but meshes only for some... I'll check what it looks like on the test atlas and update. Will add directly to master and close this PR.

vigji:
No, but probably makes sense to have what you described in the next version of the Example, to support more thorough testing

FedeClaudi:
ok, for now I've edited the tests to use what's in the example

FedeClaudi:
I commited directly to master, will close this now.

PR #65, opened 2020-07-27
Update readme

adamltyson:
I've begun to update the readme file. My aim is to provide an overview of the functionality, to persuade users to give it a go.

It is specifically not a tutorial, or an API reference, as those will come elsewhere.

I have also begun [docs.brainglobe.info](https://docs.brainglobe.info/), and the readme info is also copied there at [docs.brainglobe.info/bg-atlasapi/introduction](https://docs.brainglobe.info/bg-atlasapi/introduction). *This mirroring should be automated in the future*.

@FedeClaudi, @vigji - if you could take a look at the general structure, and let me know what you think should/should not be in it. 

**It's ended up a bit wordy,** and so some of it should be put into a tutorial notebook (could maybe be copied & pasted?). Should we have some images? Perhaps a screenshot of the GUI?


FedeClaudi:
I've only skimmed it quickly but it looks really good! 
I guess in the gitbook the usage could be in a different page from the general intro / installation and maybe we can even split things like basic usage, GUI, brain regions etc... 

Given that the docs are also on the gitbook then the README on github could be very minimal and just have a general overview/description + installation and then point to the docs for more details about usage etc. 

----

I think it'd be nice to have some images, more than happy to generate some with brainrender

----

We could have a page in the gitbook for the brainrender GUI, essentially copying what's in the README of the GUI's repo.

FedeClaudi:
Adding some comments:

- `Enhanced and Unified Mouse Brain Atlas `: the meshes are imperfect still: they're not symmetric as they should be.. 
- In `installation` you say `Python >3.8` but it works at least with `3.7` as well (that's what I use for brainrender).

Other than that the README is excellent and provides all the minimal info necessary to get people started

adamltyson:
>     * `Enhanced and Unified Mouse Brain Atlas `: the meshes are imperfect still: they're not symmetric as they should be..

I've only put this in, because it will come up when you run `brainglobe list` etc. It does work, it's just not 100%

>     * In `installation` you say `Python >3.8` but it works at least with `3.7` as well (that's what I use for brainrender).

Fixed, should say >3.8

>     * [this](https://docs.brainglobe.info/bg-atlasapi/introduction#python-api) table is not formatted nicely in the gitbook

I know, but I can't work out how to fix it!

adamltyson:
I think I'll flesh it out into seperate sections for gitbook, and then keep the "highlights" for the readme.

FedeClaudi:
> I know, but I can't work out how to fix it!

Yeah I don't know... maybe `rich` has an export as markdown function but I'm not sure. 

That sounds good, I will add the part on the GUI as a separate page if that's okay

adamltyson:
> That sounds good, I will add the part on the GUI as a separate page if that's okay

Yeah I think the gitbook should have a "group" per repo.

adamltyson:
I've cut lots from the readme, added an image and better organised the gitbook. The docs are obviously a work in progress, but is there anything we need to add right now?

vigji:
I guess this can be merged right?

adamltyson:
Yep

PR #74, opened 2020-09-03
Final touch vor v0.5

vigji:
- Moved atlases orientations to asr
- Fixed tests consequently
- Bumped minimum major version required for atlases


vigji:
I would merge it immediately, was branching just in case tests were messing up with new atlases

PR #75, opened 2020-09-03
Version 0.2.0 does not work with bg-space 0.3.0

orena1:
I don't know exactly what is the minimum version of bg-space so I set it to above 0.5.0. 
This is important for people who upgrade.


vigji:
Thanks for catching this! Yes, it is actually 0.5.0

PR #77, opened 2020-09-30
Hierarchy improvements

vigji:
Should be straightforward:
 - I added a useful method that was missing;
 - I improved the tutorial notebook (we should really generate a markdown file out of it) discussing hierarchy;
 - I removed a raised error that was preventing custom atlases in .brainglobe folder to be used.


FedeClaudi (approved):

adamltyson (approved):

adamltyson:
Looks good, but I wonder if the tutorial should have some even simpler parts to illustrate the hierarchy. 

I should know this, but do we have any syntax similar to this?

```python
structure = bg_atlas.structure_from_coords((5000, 4000, 3000))
structure.parent()
```

Maybe if the use of `get_structure_mask()`  could also be added to the tutorial?

vigji:
> Looks good, but I wonder if the tutorial should have some even simpler parts to illustrate the hierarchy.

What would you like to see there?

> I should know this, but do we have any syntax similar to this?
> 
> ```python
> structure = bg_atlas.structure_from_coords((5000, 4000, 3000))
> structure.parent()
> ```

No, because we opted for a very simple representation of the structure (no Structure objects) as I keep forgetting how to use objects when I have many around (is it a property? or an attribute? or a method? Where do I get the id? How do I get the structure dictionary? etc). But this is only my opinion, happy to hear other bells here.

> Maybe if the use of `get_structure_mask()` could also be added to the tutorial?

What do you mean? Part 1.3 is about this

PR #87, opened 2020-10-07
Updating pbar and a couple other things

FedeClaudi:


FedeClaudi:
@vigji  opened a cleaner PR here, if the tests pass we can merge?

FedeClaudi:
I've decided not to include pycln at the end because it would sometimes not work nicely with the other pre-commit checks and it was more bother than benefit...

vigji (approved):
Good for me as soon as tests are fixed!

FedeClaudi:
It seems that the linting is failing, but both black and flake8 pass on my machine. 
Any idea how to fix?

vigji:
uhm, trying updating black? I was also having something funny with
black-ing of that file...

adamltyson:
Yeah, I think it's a black version issue. We should probably update the version in  `.pre-commit-config.yaml` too.

FedeClaudi:
Done, now black did some stuff to list_atlases, le's see if that fixed it

FedeClaudi:
That fixed it, merging now.

PR #91, opened 2020-10-21
Atlas metadata print out

FedeClaudi:
Okay so I've added a `__rich_console__` method to the `BGAtlas` class, that's what's used by rich's `Console` for printing objects. If users use the builtin `print` function than `__str__` will be used instead, which prints the same but without colors. 

This is what it looks like:
image

and with the normal print
image

and `__repr__` returns: 
image

### DIFF
--- a/bg_atlasapi/bg_atlas.py
+++ b/bg_atlasapi/bg_atlas.py
@@ -186,15 +186,12 @@ class BrainGlobeAtlas(core.Atlas):
             return False
         return True
 
-    # def __repr__(self):
-    #     """Fancy print for the atlas providing authors information."""
-    #     meta = self.metadata
-    #     name_split = self.atlas_name.split("_")
-    #     pretty_name = "{} {} atlas (res. {})".format(*name_split)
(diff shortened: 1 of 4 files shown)


nickdelgrosso:
Wow, looks great!

adamltyson (approved):

vigji (commented):
Nice, looks good to me!

The only things I would raise is that 
1. Probably we want this to go in the atlas Core class, for it to be available also for custom atlases from local data - all info should be available;
2. Maybe it's a bit lot of "cosmetic" code to go in the main data-centric class, and could be refactored out in an utility function; but this is entirely subjective, and I'm happy to keep it this way and maybe move it later, it's a really minor change.

They're beautiful!

FedeClaudi:
Sure, I can move it to a function no problem

FedeClaudi:
aight, moved the `__` methods to `Core` and the bulk of the `rich` code to a utility function

vigji (approved):
Great! Good for me to merge, thanks for this!

PR #101, opened 2021-04-11
List all available atlases and their versions in machine-readable format (2)

nickdelgrosso:
Hi,

I was looking for a way to put a list of all the brainglobe atlases (as strings, to pass to the BrainGlobeAtlas constructor), and added the function by refactoring out a step in the get_local_atlas_version() function.  It works like so:

```python
>> from bg_atlasapi.list_atlases import get_all_atlases_lastversions
>> get_all_atlases_lastversions()
{ "mouse_allen_25um": "1.0",
 "rat_allen_100um": "1.2",
  ...
}
```

I've put this into slicereg, but it'd be great to put something like this upstream.  If there is a better way to get this information, though, or one I missed, I'd be happy to use it!

Best wishes,

Nick

### DIFF
--- a/bg_atlasapi/list_atlases.py
+++ b/bg_atlasapi/list_atlases.py
@@ -52,6 +52,15 @@ def get_local_atlas_version(atlas_name):
     ][0]
 
 
+def get_all_atlases_lastversions():
+    """Read from URL all available last versions"""
+    available_atlases = utils.conf_from_url(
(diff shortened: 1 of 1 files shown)


adamltyson:
Looks fine to me. @vigji @FedeClaudi?

Thanks @nickdelgrosso!

FedeClaudi:
yep

PR #102, opened 2021-04-13
Fix linting problems, and test on PR

adamltyson:


adamltyson:
This is super simple, so I'll merge. Just making sure that tests are run on PRs from other repos.

PR #106, opened 2021-11-17
Fix NumPy future warning on get structure mask

seankmartin:
Thanks so much for the package, it is great!
One small thing I noticed on the latest version:

In core.py line 349, the boolean-valued array was wrapped inside another array.
I may have missed a reason for this, but the extra array brackets [] shouldn't be needed.
This line throws a future warning on the latest version of NumPy.
The new code also matches the code on line 348 for checking descendants.

### DIFF
--- a/bg_atlasapi/core.py
+++ b/bg_atlasapi/core.py
@@ -346,7 +346,7 @@ class Atlas:
         for descendant in descendants:
             descendant_id = self.structures[descendant]["id"]
             mask_stack[self.annotation == descendant_id] = structure_id
-        mask_stack[[self.annotation == structure_id]] = structure_id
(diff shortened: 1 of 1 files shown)


adamltyson:
Thanks @seankmartin!

@vigji, @FedeClaudi was there any reason for this syntax originally?

vigji:
Thanks for the PR! I can't see a good reason for keeping the previous structure, happy to merge it!

adamltyson:
@FedeClaudi happy to merge?

FedeClaudi:
Yep, good to merge.
If there was a reason, I'm sure it wasn't a great one!

adamltyson:
Great, thanks @seankmartin! Merging.

PR #108, opened 2021-12-01
moved print to BrainGlobeAtlas

vigji:
Printing require an atlas name, which is enforced only in the BrainGlobeAtlas and does not necessarily fit requirements for a folder of a locally packaged atlas. I just moved the printing-related code to the child class. Nothing major.

### DIFF
--- a/bg_atlasapi/bg_atlas.py
+++ b/bg_atlasapi/bg_atlas.py
@@ -2,6 +2,10 @@ from pathlib import Path
 import tarfile
 import requests
 from rich import print as rprint
+from rich.console import Console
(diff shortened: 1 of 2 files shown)


FedeClaudi (approved):

FedeClaudi:
Looks good. Rich now has a [rich repr protocol](https://rich.readthedocs.io/en/latest/pretty.html?highlight=__rich_repr__#rich-repr-protocol) for even fancier `repr`, but it's probably not necessary here.

Thanks for fixing this

adamltyson (approved):

Issue #119, opened 2022-05-20
`get_structure_mask` is very slow for structures with many children


sdiebolt:
With the Allen mouse brain atlas 25µm, a call to `atlas.get_structure_mask(997)` takes approximately 45 seconds on my computer to get the whole brain mask.

Looking at `get_structure_mask`'s definition, its documentation warns that the function isn't optimized. Optimizing it could be very easy: instead of gettting each descendant's mask one by one in a loop, we could simply get the list of descendant IDs and use `numpy.isin` to get the resulting mask.

Replacing `get_structure_mask` by 

```python
structure_id = atlas.structures[structure]["id"]
descendant_ids = [
    atlas.structures[descendant]["id"] 
    for descendant in atlas.get_structure_descendants(structure_id)
]
descendant_ids.append(structure_id)

mask_stack[np.isin(atlas.annotation, descendant_ids)] = structure_id
```

leads to a big improvement:  `atlas.get_structure_mask(997)` now runs in 4 seconds.

I'd be happy to submit a pull request if you're interested!

vigji:
Sounds very reasonable! If you can that would be very welcome :) thanks for improving this!

PR #120, opened 2022-05-20
Optimize `get_structure_mask` for structures with many children

sdiebolt:
`get_structure_mask` can be optimized for structures with many children: see #119.

Thanks so much for creating and maintaining this awesome package! I'm not yet used to contributions on GitHub, feel free to let me know if I can improve anything.

### DIFF
--- a/bg_atlasapi/core.py
+++ b/bg_atlasapi/core.py
@@ -293,8 +293,9 @@ class Atlas:
 
     def get_structure_mask(self, structure):
         """Returns a stack with the mask for a specific structure (including all
-        sub-structures). This function is not particularly optimized, and might
-        take some hundreds of ms per run for some structures.
+        sub-structures).
+
+        This function might take a few seconds for structures with many children.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


vigji:
@FedeClaudi @adamltyson tests are running fine locally on my machine, but there seems to be something funky in GitHub actions going on - would you agree to merge and fix tests in a separate PR?

FedeClaudi:
I'm happy with it

sdiebolt:
It seems like the fail comes from the dependency resolver: it should not have anything to do with the changes from this PR. 

You might also want to try running the tests again: as the dependencies didn't change since last commit, it might either be a one-time bug from `pip`, or a problem introduced by a new version of one of the dependencies (in which case you might want to try to be more strict in `requirements.txt`?).

sdiebolt:
Still there after a rerun... I just tried running `pip install .[dev]` on my machine with Python 3.8.12 (as defined in the GitHub Actions workflow) and there is indeed a lot of [backtracking](https://pip.pypa.io/en/stable/topics/dependency-resolution/#possible-ways-to-reduce-backtracking). I think this could be solved very simply by setting stricter version limits in `requirements.txt`!

Changing `requirements.txt` is best left to someone who is more aware of the exact use of each of those dependencies in BG-AtlasAPI, but do tell me if I can help. :slightly_smiling_face:

vigji:
Yes it's definitively not on you @sdiebolt to fix! I'll try to give it a shot in the next days

adamltyson:
Having looked at this again, sorry @sdiebolt, you're right, the tests failing is nothing at all to do with this PR, we should merge this now, and fix the test failure separately.

sdiebolt:
No problem, that is indeed a weird test failure!

adamltyson:
@vigji are we good to merge? I haven't tested this myself yet.

PR #124, opened 2022-06-23
Fix the `-` in requirements

paddyroddy:
This currently breaks `pip install -e .[dev]`

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -3,6 +3,6 @@ coverage
 flake8
 gitpython
 isort
-pre - commit
+pre-commit
 pytest
-pytest - cov
+pytest-cov


adamltyson:
This seems to have been introduced in https://github.com/brainglobe/bg-atlasapi/pull/122, so I guess it was caused by one of the autoformatting tools.

paddyroddy:
> This seems to have been introduced in #122, so I guess it was caused by one of the autoformatting tools.

I've just ran `pre-commit` and nothing changed

paddyroddy:
My guess would be a find all and replace in an IDE and not realising what it changed. I've done this before

adamltyson (approved):

PR #127, opened 2022-07-27
Align with other BG packages

adamltyson:
A bit of a catch all PR, I've made some changes to reduce the maintenance burden by making this package more like the other BG packages. Including:

* Updating tox configuration
* Moving the requirements to the setup file. Not sure if this is good or bad, but it's what we have with every other package, and having the separate files was stopping tox from working for some reason. 
* Updated links in setup.py
* Added & updated badges in the readme
* Moved from coveralls to codecov to report coverage
* Updated supported Python versions
* Removed changelog. Although we should have one, we haven't updated this for two years, so I think it's a net negative currently.

There are no changes to the code or the packaging.


vigji (commented):
I am not a great fan of requirements embedded in the setup.py but for no strong reason, I am happy with consistency!

Is there a specific reason not to support `v3.7`?

adamltyson:
> I am not a great fan of requirements embedded in the setup.py but for no strong reason, I am happy with consistency!

That was my logic, there are a lot of repos now, and the more similar they are, the easier they are to maintain.

> Is there a specific reason not to support v3.7?

To follow [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html). Since numpy 1.22, only Python >= 3.8 are supported.

yoda-vid:
> To follow [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html). Since numpy 1.22, only Python >= 3.8 are supported.

Thanks for pointing out this link! Didn't know about this deprecation schedule, good to know.

vigji (approved):
Ok than, thanks for pointing it out, I was also not aware of it!
Good to go for me then.

PR #144, opened 2023-03-20
Build wheel when building package

dstansby:
- Use NI actions to build the package
- Add the option to test package building in pull requests by labelling as "Test packaging". Very open to changing this label name if anyone can suggest something better!


alessandrofelder (approved):
LGTM

All alternative label texts I can think of are too verbose.

PR #190, opened 2024-01-09
Various housekeeping

adamltyson:
Updates project structure, removing unused configs and updating supported Python versions.


bg_atlasapi/__init__.py
```diff
@@ -1,5 +1,13 @@
...
+from importlib.metadata import PackageNotFoundError, version
+
+try:
+    __version__ = version("bg-atlasapi")
+    del version
+except PackageNotFoundError:
+    # package is not installed
+    pass
+
+__author__ = "BrainGlobe Developers"
```
willGraham01, line 10:
```suggestion
from importlib.metadata import PackageNotFoundError, metadata

try:
    __version__ = metadata("bg-atlasapi")["Version"]
    __author__ = metadata("bg-atlasapi")["Author"]
    del metadata
except PackageNotFoundError:
    # package is not installed
    pass
```

If you want to avoid having to update the authors (and other info that is pulled from pyproject too)

willGraham01 (approved):

bg_atlasapi/__init__.py
adamltyson: Cool, thanks.

PR #191, opened 2024-01-24
Bump to brainglobe space

willGraham01:
See https://github.com/brainglobe/brainglobe-space/issues/37, new version will be `1.0.3`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ pprint(VISp)
 
 ### Note on coordinates in `bg-atlasapi`
 Working with both image coordinates and cartesian coordinates in the same space can be confusing! In `bg-atlasapi`, the origin is always assumed to be in the upper left corner of the image (sectioning along the first dimension), the "ij" convention. This means that when plotting meshes and points using cartesian systems, you might encounter confusing behaviors coming from the fact that in cartesian plots one axis is inverted with respect to  ij coordinates (vertical axis increases going up, image row indexes increase going down). To make things as consistent as possible, in `bg-atlasapi` the 0 of the meshes coordinates is assumed to coincide with the 0 index of the images stack, and meshes coordinates increase following the direction stack indexes increase.
-To deal with transformations between your data space and `bg-atlasapi`, you might find helpful the [bg-space](https://github.com/brainglobe/bg-space) package.
+To deal with transformations between your data space and `bg-atlasapi`, you might find helpful the [brainglobe-space](https://github.com/brainglobe/brainglobe-space) package.
 
 # Contributing to bg-atlasapi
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -128,7 +128,7 @@ pprint(VISp)
 
 ### Note on coordinates in `bg-atlasapi`
 Working with both image coordinates and cartesian coordinates in the same space can be confusing! In `bg-atlasapi`, the origin is always assumed to be in the upper left corner of the image (sectioning along the first dimension), the "ij" convention. This means that when plotting meshes and points using cartesian systems, you might encounter confusing behaviors coming from the fact that in cartesian plots one axis is inverted with respect to  ij coordinates (vertical axis increases going up, image row indexes increase going down). To make things as consistent as possible, in `bg-atlasapi` the 0 of the meshes coordinates is assumed to coincide with the 0 index of the images stack, and meshes coordinates increase following the direction stack indexes increase.
-To deal with transformations between your data space and `bg-atlasapi`, you might find helpful the [bg-space](https://github.com/brainglobe/bg-space) package.
+To deal with transformations between your data space and `bg-atlasapi`, you might find helpful the [brainglobe-space](https://github.com/brainglobe/brainglobe-space) package.
```
alessandrofelder, line 131:
```suggestion
To deal with transformations between your data space and `bg-atlasapi`, you might find the [brainglobe-space](https://github.com/brainglobe/brainglobe-space) package helpful.
```
better grammar?

pyproject.toml
```diff
@@ -108,5 +95,5 @@ python =
 [testenv]
 extras =
     dev
-commands = pytest -v --color=yes --cov=bg_atlasapi --cov-report=xml
+commands = pytest -v --color=yes --cov=bg_atlasapi --cov-report=xml -W ignore::DeprecationWarning
```
alessandrofelder, line 98: Just for my understanding. Why do we decide to additionally add ignoring a DeprecationWarning here - is something going to be broken or is this an auto-formatting fix?

alessandrofelder (approved):

pyproject.toml
willGraham01: pandas is throwing a depreciation warning that causes our CI tests to fail (see https://github.com/brainglobe/bg-atlasapi/actions/runs/7640045487/job/20814376965#step:2:271). It's a futurewarning, but because our tox environment is set to elevate warnings to errors, CI just fails.

alessandrofelder: Thanks for explaining.
Should we add PyArrow as a dependency instead to avoid breakage when the new major Pandas release happens? (Possibly in a different issue + PR?) The point of having the tox environment set up like that is to fix these things early, I think?

PR #192, opened 2024-01-25
Add CITATION.cff

willGraham01:
Related to https://github.com/brainglobe/BrainGlobe/issues/61, this update adds a `CITATION.cff` file to the repository. `.cff` format [is explained and can be generated here](https://citation-file-format.github.io/).

Amongst other things, this will mean that [GitHub will provide](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files) citation information (and source code, should we want it!) when visiting the repository homepage.

Moreover, it moves us towards the objective of https://github.com/brainglobe/BrainGlobe/issues/61 by adding a custom `citation-sentence` field. This is intended for our own internal use in a `brainglobe-meta` `cite()` function, which can fetch the `.cff` files that are in our repositories.

(If #191 goes in before this, this will need an update to the latest version and corresponding hash. It also depends on #191 implicitly, since this fixes the PANDAS warning that we are seeing in CI).


alessandrofelder (approved):
Looks good to me.

It might be worth making an (meta)issue for @adamltyson to double-check all the cff files once they exist, just to be sure?

> If https://github.com/brainglobe/bg-atlasapi/pull/191 goes in before this, this will need an update to the latest version and corresponding hash.

I can't tell if you've done this but let this be a reminder that you think you should, if you haven't yet.

willGraham01:
I'll update https://github.com/brainglobe/BrainGlobe/issues/61 with a list of our repos that we add CITATION.cffs to. Once the citation functionality goes into `utils` we can then scan through them all, and can add preferred citations if needs be

PR #299, opened 2024-05-17
Return "Outside atlas" rather than key error

### LINKED ISSUE #157: [Feature] Add an "Out of brain" acronym for the ID 0
When querying for structures as ID from coordinates, out of bounds coordinates return 0, but when asking for an acronym, it raises a KeyError value.

Add an acronym to the id 0, named for example "OutOfBrain" or whatever.

**Describe alternatives you've considered**
This can be managed when using the API with a try/catch but I guess it would be nicer if it is handled directly in the bg-atlasapi.

**Additional context**
Code to reproduce the error : 
```
from bg_atlasapi import BrainGlobeAtlas

atlas = BrainGlobeAtlas("kim_mouse_10um", check_latest=False)  # does the same thing with allen_mouse_10um
atlas.structure_from_coords((9386.32578694152, 6137.89857535182, 8869.46676429553), microns=True) # returns 0
atlas.structure_from_coords((9386.32578694152, 6137.89857535182, 8869.46676429553), microns=True, as_acronym=True) # raises KeyError: 0
```

adamltyson:
## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**
When querying for structures as ID from coordinates, out of bounds coordinates return 0, but when asking for an acronym, it raises a KeyError value.

**What does this PR do?**
Rather than a key error, "Outside atlas" is returned

## References

Closes #157

New tests added & checked manually

## Is this a breaking change?

Nope

Don't think so

## Checklist:

### DIFF
--- a/tests/atlasapi/test_core_atlas.py
+++ b/tests/atlasapi/test_core_atlas.py
@@ -102,6 +102,19 @@ def test_data_from_coords(atlas, coords):
     )
 
 
+def test_data_from_coords_out_of_brain(atlas, coords=(1, 1, 1)):
+    assert atlas.structure_from_coords(coords) == 0
+    assert atlas.structure_from_coords(coords, microns=True) == 0
+
+    assert (
+        atlas.structure_from_coords(coords, as_acronym=True) == "Out of brain"
+    )
+    assert (
+        atlas.structure_from_coords(coords, microns=True, as_acronym=True)
+        == "Out of brain"
+    )
+
+
 def test_meshfile_from_id(atlas):
     assert (
         atlas.meshfile_from_structure("CH")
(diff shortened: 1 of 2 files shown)


tests/atlasapi/test_core_atlas.py
```diff
@@ -102,6 +102,22 @@ def test_data_from_coords(atlas, coords):
     )
 
 
+def test_data_from_coords_out_of_brain(
+    atlas, coords=(1, 1, 1), key_error_string="Outside atlas"
```
alessandrofelder, line 106: What's the reason for `coords` and `key_error_string` to be arguments of the test function?

I think they should just be defined in the test function body?

alessandrofelder (approved):
Minor style points, otherwise LGTM

tests/atlasapi/test_core_atlas.py
adamltyson: No reason. I think I usually do it so anything that may need to be changed is at the top of the function. Will move it into the function.

PR #300, opened 2024-05-17
Remove slow test marker

### LINKED ISSUE #297: Ensure all tests are run
There is currently [one test](https://github.com/brainglobe/brainglobe-atlasapi/blob/23abdcca41b5c30368a409cd1caa66ef4e5287af/tests/atlasapi/test_config_file.py#L34) skipped by default. This test should either be made to run all the time, or not at all.

adamltyson:
This PR closes #297 by removing the slow pytest marker.

I changed the test a bit, but it's still not ideal, because it changes the default brainglobe config file. It changes it back, but that's not great.

I assumed I could get around this by setting `monkeypatch.setenv("BRAINGLOBE_CONFIG_DIR", str(tmp_path))`, this however doesn't seem to set the config dir temporarily. I don't know if I'm monkeypatching this incorrectly, or there's a bug in the atlas api itself.

I've created this as a draft PR because I think this should be fixed, but maybe this PR should be merged and then an issue created to improve the test? This PR doesn't introduce the problem.

### DIFF
--- a/tests/atlasapi/test_config_file.py
+++ b/tests/atlasapi/test_config_file.py
@@ -26,33 +26,33 @@ def test_config_creation(conf_path):
             assert val == str(config.TEMPLATE_CONF_DICT[sectname][k])
 
 
-# Ugly test zone: here we use the terminal commands, which edit the config
-# file in the brainglobe_atlasapi repo from which the tests are being run.
-# This is not the cleanest way, the alternative would be to run this test in
-# a new env.
-@pytest.mark.slow
-def test_config_edit():
+def test_config_edit(tmp_path):
+    # Create a CliRunner instance
     runner = CliRunner()
     result = runner.invoke(cli.bg_cli, ["config", "--show"])
+
+    # Assert the command ran successfully
     assert result.exit_code == 0
     assert result.output == config._print_config() + "\n"
 
+    # Read the current config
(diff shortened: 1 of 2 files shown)


alessandrofelder:
>I've created this as a draft PR because I think this should be fixed, but maybe this PR should be merged and then an issue created to improve the test? This PR doesn't introduce the problem.

I don't mind either way. A new issue is probably cleaner (but also more overhead). One thing maybe worth trying (that I think would be a good idea across BrainGlobe) is to [monkeypatch `Path.home()` for all tests when not running on CI.](https://github.com/brainglobe/brainrender-napari/blob/9b3253758ea9efc4c56bb79d060d0af8c058420f/tests/conftest.py#L11) It looks like there's a few constant variables that need to be adapted for the patching to work. This way
* we test a slightly more real-life case on CI (`Path.home()` is not patched, but that's fine, because the CI does not have a user set up.)
* we locally keep `~./brainglobe/` (user data) and `~/.brainglobe-tests/.brainglobe/` (tests interacting with fake user data) always completely separate.

adamltyson:
To simplify, I've marked this ready to review and raised https://github.com/brainglobe/brainglobe-atlasapi/issues/302

alessandrofelder (approved):

PR #290, opened 2024-05-14
Fixing human atlas script

viktorpm:
## Description

**What is this PR?**

- [x] Bug fix
- [ ] Addition of a new feature
- [ ] Other

**What does this PR do?**

- Fixes `humanatlas.py` atlas generation script. 
- Automates the annotation and reference downloading steps.
- Fixes #275

## How has this PR been tested?

- locally
- on the HPC
- the regenerated atlas was validated

## Is this a breaking change?

no

## Does this PR require an update to the documentation?

no

## Checklist:

- [x] The code has been tested locally
- [ ] Tests have been added to cover all new functionality (unit & integration)
- [ ] The documentation has been updated to reflect any changes
- [x] The code has been formatted with [pre-commit](https://pre-commit.com/)

PR #305, opened 2024-05-21
Fix AZBA

### LINKED ISSUE #213: [Feature] Force all atlas 3D meshes to be in consistent "brainglobe" space, aligned to the annotations
The annotation volumes are set up so that the origin (0,0,0) is at the front/left/top corner. Some of the 3D mesh files are aligned with this so that the origin in 3D space is also at this same point (e.g. see the CCF files), but many of the atlases are not (see e.g. the human atlas or AZBA). It's possible to apply the `trasform_to_brainglobe` affine transform to move the brains to the correct position in some cases, but in other cases the transform appears to be incorrect.

It would be a big benefit for users to simply move all the 3D meshes into the annotation space for consistency (i.e. so that the transform is the identity matrix with a (0,0,0) translation)

adamltyson:
Fixes the adult zebrafish atlas. The meshes were scaled correctly, so it just needed the parameter passing to the wrap up function. I also made some small changes, such as downloading the data with pooch. i removed the options for parallel processing, because it always seems to crash (as seen with other scripts). 

There's still a lot that can be done to tidy up and standardise things, but I thought I'd raise this PR now as it seems to fix the issue (need to run the script through fully first to be sure though). I guess it's worth going through and standardising all the atlas scripts in one go @alessandrofelder, @viktorpm?

@alessandrofelder one thing that I don't understand - the meshes look fine in brainrender, but in brainrender-napari, they are basically invisible. The easiest way to know they are there is because they make the underlying arrays look darker. I haven't seen this before, do you have any idea?

We're going through all the atlases to fix any problems, but #213 specifically mentions the human atlas and this one. Following #290, I guess this PR closes #213.

### DIFF
--- a/brainglobe_atlasapi/atlas_generation/atlas_scripts/azba_zfish.py
+++ b/brainglobe_atlasapi/atlas_generation/atlas_scripts/azba_zfish.py
@@ -79,17 +79,15 @@ def create_atlas(working_dir, resolution):
     # 'id', 'structure_id_path', and 'rgb_triplet' key values
     for i in range(0, len(hierarchy)):
         hierarchy[i]["id"] = int(hierarchy[i]["id"])
-    for j in range(0, len(hierarchy)):
-        hierarchy[j]["structure_id_path"] = list(
-            map(int, hierarchy[j]["structure_id_path"].split("/"))
+        hierarchy[i]["structure_id_path"] = list(
+            map(int, hierarchy[i]["structure_id_path"].split("/"))
         )
-    for k in range(0, len(hierarchy)):
         try:
-            hierarchy[k]["rgb_triplet"] = list(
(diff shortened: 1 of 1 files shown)


adamltyson:
The packaged atlas looks fine, and passes the validation script. It just looks weird in brainrender-napari. The problem seems to be specific to the root mesh though:
azba

viktorpm:
Yes, there's already a draft PR for standardisation: #296

adamltyson:
From AF: the RGB triplet for the root mesh may be black, should update it to white.

alessandrofelder (approved):
Apart from the potential RGB issue, looks good to me.

adamltyson:
RGB issue fixed, atlas passes validation script and works well in brainrender, brainrender-napari & brainreg.

PR #310, opened 2024-06-14
Move `pymcubes` to an optional dependency.

willGraham01:
`PyMCubes` is a dependency that is only required during atlas generation, and [currently doesn't have an ARM package](https://github.com/brainglobe/brainreg/issues/206#issuecomment-2167579703) on conda-forge.

Forcing it to be a core dependency results in failed builds when installing via conda-forge, and since it's not needed for registration, we have decided to move it to an optional `atlasgen` dependency that users will not obtain by default.

After this, we'll need a new PyPI release and then to fix the corresponding conda-forge feedstock PR that will be created.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,6 @@ dependencies = [
     "numpy",
     "pandas",
     "pyarrow",
-    "PyMCubes",
     "requests",
     "rich >= 9.0.0",
     "SimpleITK",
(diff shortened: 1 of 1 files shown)


adamltyson:
Should we add other packages that aren't needed for the atlasAPI? Maybe ITK is one?

willGraham01:
A quick search of this repo indicates that the following packages are only used in the `atlas_generation/` submodule:

- loguru
- PyMCubes
- SimpleITK
- tqdm
- vedo
- xmltodict

Also, we seem to be fetching `scikit-image` but never actually importing it... so I guess it can just go?

alessandrofelder (approved):
Looks good to me. Would you mind open an issue in the docs repo to update the "how to contribute an atlas" accordingly, please?

willGraham01:
https://github.com/brainglobe/brainglobe.github.io/issues/203

PR #328, opened 2024-07-22
Ensure no allen_cord meshes are black

adamltyson:
The updated atlas script looked fine, but some meshes were black, so they didn't show up in `brainrender-napari`.


adamltyson:
Test failures are weird, but unrelated.

alessandrofelder (approved):

Issue #206, opened 2023-11-11
Add cuttlefish atlas


adamltyson:
https://www.cuttlebase.org/downloads

alessandrofelder:
Noting that we have a student working on this ATM, to avoid duplication of efforts :)

adamltyson:
#206 adds the brain atlas. Once merged, lets keep this open to track potentially adding the full body atlas. cc @PolarBean

PR #338, opened 2024-08-02
Add Cuttlebase cuttlefish atlas

kjungwoo5:
## Description

**What is this PR**

- [ ] Bug fix
- [X] Addition of a new feature
- [ ] Other

**Why is this PR needed?**
We would like to add support for the cuttlebase.org cuttlefish atlas. 

**What does this PR do?**
This adds a packaging script for the cuttlefish atlas. 

## References
Issue #206 

The code has been tested locally. 

## Is this a breaking change?
No. 

If any features have changed, or have been added. Please explain how the documentation has been updated (and link to the associated PR). See [here](https://brainglobe.info/community/developers/index.html#to-improve-the-documentation) for details.

## Checklist:

- [X] The code has been tested locally


kjungwoo5:
Hi @alessandrofelder, I think the code is ready for review now. Thank you!

brainglobe_atlasapi/atlas_generation/atlas_scripts/cuttlefish.py
```diff
@@ -0,0 +1,357 @@
...
+    # import cuttlefish .nii file
+    template_path = pooch.retrieve(
+        TEMPLATE_URL,
+        known_hash="195125305a11abe6786be1b32830a8aed1bc8f68948ad53fa84bf74efe7cbe9c",  # noqa E501
+        progressbar=True,
+    )
+
+    # process brain template MRI file
+    print("Processing brain template:")
+    brain_template = load.load_nii(template_path, as_array=True)
```
alessandrofelder, line 253: We need to rescale the template image here, because otherwise it get naively converted from `np.float64` to `np.uint16`. For an example of how this is done, see https://github.com/brainglobe/brainglobe-atlasapi/blob/a1cbccc1369a57762878ef801167d214a007e229/brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_devccf_mouse.py#L94 Lines 94-99

alessandrofelder (commented):
Thanks @kjungwoo5 - one last hurdle!

alessandrofelder:
> Thanks @kjungwoo5 - one last hurdle!

I spoke too soon 🤦 - sorry. There is a second final hurdle: the annotation pixel value of equivalent regions in left and right hemisphere has to be the same in BrainGlobe - downstream tools depend on this. So we need some code to replace the pixels in the right hemisphere with the value of the pixel of the equivalent region in the left.

PolarBean:
In addition to the brain atlas there is also a whole body cuttlefish atlas mentioned in this publication. It would be really cool to see that as another atlas available in brainglobe!
Congratulations on this pull request it looks good :)

PolarBean:
https://www.cuttlebase.org/3d-cuttlefish
image

PolarBean:
One thing I noticed is that you include  mesh url to the cuttlfish meshes but you don't use it, instead generating your own meshes from the segmentation. Is this intended?

adamltyson:
> One thing I noticed is that you include mesh url to the cuttlfish meshes but you don't use it, instead generating your own meshes from the segmentation. Is this intended?

I assume not. If there are existing meshes, we should use them.

alessandrofelder:
@kjungwoo5 and I tried very hard to use the existing meshes. 
* They are in a format that standard Python mesh reading libraries can't read, so we made our own reader. 
* Then we noticed that they were in a different coordinate to the image, and the transformation was undocumented, as far as we could see. 
* So we decided to reduce the scope of v1 of the cuttlefish atlas to recompute the meshes ourselves for now.

PolarBean:
What an effort! Just goes to show the value of standardising all components of atlases :)

brainglobe_atlasapi/atlas_generation/atlas_scripts/cuttlefish.py
kjungwoo5: Sounds good, I'll try to get working on this soon!

kjungwoo5:
> > Thanks @kjungwoo5 - one last hurdle!
> 
> I spoke too soon 🤦 - sorry. There is a second final hurdle: the annotation pixel value of equivalent regions in left and right hemisphere has to be the same in BrainGlobe - downstream tools depend on this. So we need some code to replace the pixels in the right hemisphere with the value of the pixel of the equivalent region in the left. I can help with this, if you like, @kjungwoo5 ?

I would appreciate your help on this, thank you!

kjungwoo5:
> One thing I noticed is that you include mesh url to the cuttlfish meshes but you don't use it, instead generating your own meshes from the segmentation. Is this intended?

Ah yes, thank you for pointing that out. That is a remnant of our previous attempts to use the existing meshes, as per @alessandrofelder's explanation. I will remove it in my next commit.

PolarBean:
Hey I made a commit to update the script name to match the api name :) Let me know if this is the right way to do this. I'm not sure how to use the github propose changes feature,.

alessandrofelder:
@kjungwoo5 I had some time to work on symmetrising the annotations for this atlas.
Can I ask some clarifying questions:
- what were the `hierarchy[-3]` and `hierarchy[-4]` that had an extra key? I didn't run into this problem after my modifications (maybe they were in the right hemisphere, and I'm only keeping the left)?
- what allowed you to deduce that we needed additional ids for SB and IB?
- remind me what you mean by dummy values in your [comment on line 166](https://github.com/kjungwoo5/brainglobe-atlasapi/blob/8cf0d381549eb65dbfae38feca69e9005214dcab/brainglobe_atlasapi/atlas_generation/atlas_scripts/columbia_cuttlefish.py#L166)

Thanks :)

kjungwoo5:
> @kjungwoo5 I had some time to work on symmetrising the annotations for this atlas. Can I ask some clarifying questions:
> 
> * what were the `hierarchy[-3]` and `hierarchy[-4]` that had an extra key? I didn't run into this problem after my modifications (maybe they were in the right hemisphere, and I'm only keeping the left)?
> * what allowed you to deduce that we needed additional ids for SB and IB?
> * remind me what you mean by dummy values in your [comment on line 166](https://github.com/kjungwoo5/brainglobe-atlasapi/blob/8cf0d381549eb65dbfae38feca69e9005214dcab/brainglobe_atlasapi/atlas_generation/atlas_scripts/columbia_cuttlefish.py#L166)
> 
> Thanks :)

Hi Alessandro, sorry for the delay in getting back to you! 

* `hierarchy[-3]` is the Visceral Nerves (Right), and `hierarchy[-4]` is the Visceral Nerves (Left). The issue is that the functions of these regions were written in a full sentence containing commas, so when we try to read the whole hierarchy file, these two regions were given extra dictionaries with a key of `None`, and values containing fragments of the original sentence. 
* I believe it's because if I left them to follow my code, they would have been assigned the IDs 30 and 31 respectively, which was already given to ADCl and ADCr respectively, so I included these 2 cases in order to prevent different regions from having the same ID. 
* The dummy values refer to the `structure_id_path` used by the cuttlebase team, which represented the tree for the regions. However, as they did not use unique ID values in the tree, it was unusable for our code, so I replaced the values they used within their `structure_id_path` with the correct corresponding IDs which we had generated earlier in the code 

I hope this explains it, and please let me know if any of this was unclear!

alessandrofelder (approved):
I've fixed the symmetry issue - it now all looks reasonable and valid according to our validation functionality and manual inspection. I will therefore merge this and upload the new atlas (I'm not sure I understand the data and the code entirely, but I am calling this good enough now).

Issue #203, opened 2023-11-20
Support the Gubra "Multimodal 3D Mouse Brain Atlas Framework with the Skull-Derived Coordinate System "


adamltyson:
https://doi.org/10.1007/s12021-023-09623-9

adamltyson:
URL: https://www.neuropedia.dk/resource/multimodal-3d-mouse-brain-atlas-framework-with-the-skull-derived-coordinate-system/

PolarBean:
it should be worth mentioning that this also includes a new LSFM atlas which differs significantly from the one which is currently in the brainglobe api. They seem to have an error in the label file at present but once this is resolved I will integrate it here.

Issue #284, opened 2024-05-10
Use pooch to download data


adamltyson:
Pooch could be used to download:
- The atlases as part of the API
- The data used to generate atlases

This would provide many benefits, mostly in that we could remove all of our own code to do this, and it would also provide validation that atlases are properly downloaded (with checksums etc).

adamltyson:
this should hopefully fix - https://github.com/brainglobe/brainglobe-atlasapi/issues/334

PolarBean:
I support using pooch to download the atlases in the atlas creation scripts, but one benefit of using a centralised database of atlases is we avoid link rot. So many atlasing papers I have been reviewing recently have dead links to the data files. In many cases these datasets are completely lost. Maybe this is out of scope for the API though.

PolarBean:
ah that makes sense then!

NicoKiaru:
I don't know if it's really part of this issue, but why not storing the atlas raw data on Zenodo ? I experience a very slow download speed with brainglobe (less than a 1MB/s). Zenodo has a very good infrastructure and allow direct dl links without login, etc. And I have a dl speed of over 50 Mb/s over wifi, so wifi is probably limiting here.

adamltyson:
It's a different issue, but a good point. We're looking at setting up mirrors to improve download speed, but zenodo does seem like a good option. I'm not sure what happened with GIN, I don't recall it being this slow when we first set it up.

IgorTatarnikov:
We're now using s3fs to download the data for S3. I think this issue can now be closed?

adamltyson:
Yes, I think so

PR #370, opened 2024-08-23
Perens stereotaxic mri

PolarBean:
Here I added the Perens MRI stereotaxic atlas

## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

## References

https://github.com/brainglobe/brainglobe-atlasapi/issues/203

I ran the atlas generation script, and inspected the output. 
I attempted to run the mesh checking however I am missing the correct drivers for libGL.


brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
```diff
@@ -0,0 +1,402 @@
+__version__ = "3"
```
adamltyson, line 1: Our versioning is (currently) independent from the atlas version, so this will now make version 1.0.

```suggestion
__version__ = "0"
```

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
```diff
@@ -0,0 +1,402 @@
...
+    Region,
+    create_region_mesh,
+)
+from brainglobe_atlasapi.atlas_generation.wrapup import wrapup_atlas_from_data
+from brainglobe_atlasapi.structure_tree_util import get_structures_tree
+
+PARALLEL = True  # disable parallel mesh extraction for easier debugging
+
+
+### Additional functions #####################################################
```
alessandrofelder, line 25: To avoid code duplication, I would suggest to use existing functions to map from id to acronym and other queries on the Allen atlas. Some example relevant functionality from an IPython notebook:
```
In [11]: atlas = BrainGlobeAtlas("allen_mouse_25um")

In [12]: atlas.structures.acronym_to_id_map["mp"]
Out[12]: 673

In [13]: atlas.structures["673"]
Out[13]: {'acronym': 'mp', 'id': 673, 'name': 'mammillary peduncle', 'structure_id_path': [997, 1009, 991, 824, 46, 673], 'rgb_triplet': [204, 204, 204], 'mesh_filename': PosixPath('/home/alessandro/.brainglobe/allen_mouse_25um_v1.2/meshes/673.obj'), 'mesh': None}

In [14]: atlas.structures["673"]["acronym"]
Out[14]: 'mp'

In [15]: atlas.get_structure_ancestors("673")
Out[15]: ['root', 'fiber tracts', 'mfbs', 'mfsbshy', 'mfbsma']
```

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
```diff
@@ -0,0 +1,402 @@
...
+        with (
+            urllib.request.urlopen(req) as response,
+            open(destination_path, "wb") as out_file,
+        ):
+            data = response.read()  # a `bytes` object
+            out_file.write(data)
+        with py7zr.SevenZipFile(destination_path, mode="r") as z:
+            z.extractall(path=atlas_files_dir)
+        destination_path.unlink()
+
```
alessandrofelder, line 208: `pooch` _should_ be able to take care of all of the below, and would make this considerably shorter, and more consistent with newer atlas packaging scripts (we aim to move all atlas packaging to `pooch` one day #284 ). See [`axolotl_atlas.py` for an example](https://github.com/brainglobe/brainglobe-atlasapi/blob/086ef0db7761ad3051d544d0228aa448f03246eb/brainglobe_atlasapi/atlas_generation/atlas_scripts/axolotl_atlas.py#L57C2-L64C10).

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
```diff
@@ -0,0 +1,402 @@
...
+        / "mri_ano.nii.gz"
+    )
+    reference_file = (
+        atlas_files_dir
+        / "Multimodal_mouse_brain_atlas_files"
+        / "MRI_space_oriented"
+        / "mri_temp.nii.gz"
+    )
+
+    annotated_volume = sitk.GetArrayFromImage(
```
alessandrofelder, line 228: For consistency with other packaging scripts, can we use `brainglobe_utils.IO.image.load_any` instead of ITK here and lines below?

pyproject.toml
```diff
@@ -56,6 +56,7 @@ allenmouse = ["allensdk"]
 allenmouse_barrels = [
     "allensdk",
     "voxcell"]
+perens_stereotaxic_mri_mouse = ["py7zr"]
```
alessandrofelder, line 59: hopefully we don't need an extra dependency here, if we move to `pooch`?

alessandrofelder (commented):
Thanks a lot for this @PolarBean 
I've left some high-level comments for now - let me know what you think.

pyproject.toml
PolarBean: I think we still do. I will move the script to pooch but this file is in 7zip format which I think requires the py7zr dependency to decompress.

alessandrofelder: You're right - yep, thank yoU!

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
PolarBean: Unfortunately it seems pooch cannot handle this as it requires specific headers (The gubra server has cloudflare protection which refuses connections that are not from a browser. so my workaround is to set the request headers to that of a browser). I tried to implement a similar solution in pooch with no luck.

alessandrofelder: Great detective work, thanks - apologies for leading you down that rabbit hole!
I'd suggest encapsulating this part into a function, and document why we need this instead in its docstring?

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
PolarBean: I tried to address this in the latest commit, let me know if its what you had in mind.

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
PolarBean: Great idea, done

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
PolarBean: done

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
```diff
@@ -0,0 +1,325 @@
...
+def tree_traverse_child2parent(df, child_id, ids):
+    parent = df["parent_id"][df["id"] == child_id].item()
+
+    if not np.isnan(parent):
+        id = df["id"][df["id"] == parent].item()
+        ids.append(id)
+        tree_traverse_child2parent(df, parent, ids)
+        return ids
+    else:
+        return ids
```
alessandrofelder, line 86: I think this function is now unused (and should therefore be removed)?

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
```diff
@@ -0,0 +1,325 @@
...
+
+from brainglobe_atlasapi import BrainGlobeAtlas, utils
+from brainglobe_atlasapi.atlas_generation.mesh_utils import (
+    Region,
+    create_region_mesh,
+)
+from brainglobe_atlasapi.atlas_generation.wrapup import wrapup_atlas_from_data
+from brainglobe_atlasapi.structure_tree_util import get_structures_tree
+
+atlas = BrainGlobeAtlas("allen_mouse_25um")
```
alessandrofelder, line 22: I would maybe rename this and add a comment for legibility? So we avoid confusion between the atlas packaged here and the Allen one? 
```suggestion
# The Perens atlas re-uses information from the Allen atlas, so it's useful to have an instance of the Allen atlas around
allen_atlas = BrainGlobeAtlas("allen_mouse_25um")
```

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
```diff
@@ -0,0 +1,325 @@
...
+    "Sec-Fetch-Mode": "navigate",
+    "Sec-Fetch-Site": "none",
+    "Sec-Fetch-User": "?1",
+    "TE": "trailers",
+    "Priority": "u=0, i",
+}
+
+
+def download_and_extract_files(ATLAS_FILE_URL, destination_path):
+    """This is needed to get the brainglobe data from their server,
```
alessandrofelder, line 61:
```suggestion
    """This is needed to get the source data from their server,
```

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
```diff
@@ -0,0 +1,325 @@
...
+    Call:
+        get_all_children(df, key)
+
+    Args:
+        key (int/string)      : atlas region ID/acronym
+
+    Returns:
+        parents (list) : brain region acronym corresponding to input ID
+    """
+    parents = atlas.get_structure_ancestors(temp_id)
```
alessandrofelder, line 102:
```suggestion
parents = atlas.get_structure_ancestors(key)
```
I think?

Also, not sure these two lines are worth an additional function? It's only used once, and could just be an extra line in the `for` loop to reduce levels of abstraction?

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
```diff
@@ -0,0 +1,325 @@
...
+    else:
+        return ids
+
+
+def get_all_parents(atlas, key):
+    """
+    Get all parent IDs/acronyms in Allen's brain atlas hierarchical structure'
+
+    Call:
+        get_all_children(df, key)
```
alessandrofelder, line 94:
```suggestion
        get_all_parents(df, key)
```

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
```diff
@@ -0,0 +1,325 @@
...
+        ids.append(id)
+        tree_traverse_child2parent(df, parent, ids)
+        return ids
+    else:
+        return ids
+
+
+def get_all_parents(atlas, key):
+    """
+    Get all parent IDs/acronyms in Allen's brain atlas hierarchical structure'
```
alessandrofelder, line 91: this function is not specific to the Allen, right? But as I mention below, I'd be tempted to remove this function altogether?

alessandrofelder (changes requested):
Looking good - thanks @PolarBean 
Some more very minor comments.

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
PolarBean: oops

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
PolarBean: I've instead used the bgapi to simply replicate the allen atlas structure list since they are the same. I can change this if it's not recommended.

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
```diff
@@ -0,0 +1,336 @@
...
+        """This is needed to get the source data from their server,
+        and bypass cloudflare which is only allowing browser access"""
+        req = urllib.request.Request(ATLAS_FILE_URL, headers=HEADERS)
+        with (
+            urllib.request.urlopen(req) as response,
+            open(destination_path, "wb") as out_file,
+        ):
+            data = response.read()  # a `bytes` object
+            out_file.write(data)
+        with py7zr.SevenZipFile(destination_path, mode="r") as z:
```
adamltyson, line 127: Is there a way to do this without adding a new dependency (`py7zr`)?

adamltyson:
@PolarBean is it possible to add a few more print statements, and ideally a download progress bar? It's hard to know what's going on when running this script.

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
adamltyson: Cool, just wondering.

adamltyson:
Thanks for working on this @PolarBean. 

I'm not exactly sure what files are packaged with this atlas, but presumably we should be packaging up multiple reference (template) images? If I understand correctly, the main reason to add this atlas would be to make available all the different reference images as registration targets? 

It may also be worth packaging up multiple atlases, one for each coordinate space.

PolarBean:
> Thanks for working on this @PolarBean.
> 
> I'm not exactly sure what files are packaged with this atlas, but presumably we should be packaging up multiple reference (template) images? If I understand correctly, the main reason to add this atlas would be to make available all the different reference images as registration targets?
> 
> It may also be worth packaging up multiple atlases, one for each coordinate space.

This atlas only has one template image (MRI). It is in a different space to the perens LSFM atlas. It's valuable because it is in flat skull position and is therefore useful for stereotaxic surgery. It is also useful as a MRI registration target. I will make a seperate pull request for the most recent perens lsfm volume once they have updated those files on their server.

adamltyson:
Ah, nvm. I assumed as it was a "multimodal" atlas framework, there were templates for LSFM & CT in the same space.

PolarBean:
Yeah they have multiple templates in multiple spaces and they are linked by deformation matrices 
I've also developed https://github.com/Neural-Systems-at-UIO/CCF_translator to make transferring data between the spaces more user friendly.

PolarBean:
> If I understand correctly, the main reason to add this atlas would be to make available all the different reference images as registration targets?
Also one thing I'm curious about is what should our approach be to adding new resources to brainglobe. Should a new atlas provide some benefit in order to be added, or instead is just the existence of a resource enough to add it (even if it has been superseded by newer better resources)? I would tend towards wanting to add in everything, as it could be useful for say replicating past results or comparing atlas definitions of particular regions. But others may disagree.

adamltyson:
I agree with you. As an example [this atlas](https://brainglobe.info/documentation/brainglobe-atlasapi/usage/atlas-details.html#smoothed-version-of-the-kim-et-al-mouse-reference-atlas) was added to allow for some pre-BrainGlobe analyses to be replicated with BrainGlobe tools.

If an atlas has obviously been replaced by another, then I would suggest we:
- Make sure this is noted on the website to ensure users are using the best atlas for them
- We don't prioritise adding it, and try to focus on the atlases with the biggest impact (which I assume will be the "better" ones)

PolarBean:
> @PolarBean is it possible to add a few more print statements, and ideally a download progress bar? It's hard to know what's going on when running this script.

No worries, I've added a download progress bar and a few more print statements. Let me know if it's what you had in mind

adamltyson:
> No worries, I've added a download progress bar and a few more print statements. Let me know if it's what you had in mind

Yes, thanks! A bit more printing (e..g just to say what's being downloaded might be useful too).

I haven't checked the packaged atlas in detail (I'll leave that to @alessandrofelder when he's back), but I noticed a couple of issues:
- The packaged atlas is in `ars` orientation (rather than `asr`). This would suggest that the orientation set in the script is wrong. This may be because `load_any()` is loading the raw image data, and not transforming the data according to the nifti metadata
- There are many more stuctures (840) than meshes (754). It's not always possible to create a mesh for every region, but nearly 100 missing is surprising. Not sure why this has happened though.

PolarBean:
> * The packaged atlas is in `ars` orientation (rather than `asr`). This would suggest that the orientation set in the script is wrong. This may be because `load_any()` is loading the raw image data, and not transforming the data according to the nifti metadata

I see, I'll have to recheck it. I thought it was unchanged from the previous perens lsfm in the apiwhich is listed as IAL and so I copied it from there. I will recheck it for this one and for the other pull requests I have.

PolarBean:
>     * There are many more stuctures (840) than meshes (754). It's not always possible to create a mesh for every region, but nearly 100 missing is surprising. Not sure why this has happened though.

my guess is that its to do with this atlas being derived from the allen but is lower resolution. Therefore many of the meshes which are large enough to be rendered in the allen atlas (particularly in 10um) are too small here. 

Its actually surprising that this atlas has a voxel size of 25um but contains fewer voxels than does the allen 25um. I had thought that LSFM actually expanded the brain, not shrank it. but I may be confusing things with expansion microscopy.

PolarBean:
> * There are many more stuctures (840) than meshes (754). It's not always possible to create a mesh for every region, but nearly 100 missing is surprising. Not sure why this has happened though.

Hmm, I contacted Gubra about some problems I was facing when ingesting the LSFM volume (for which I will soon open a pull request). They fixed that problem and it seems they also updated the MRI annotation while they were at it. When I run this script I now get 838 meshes which seems much more respectable!

PolarBean:
I read the documentation but Im a bit confused about differentiating between ars and als for symmetrical atlases. I put als as i thought left to right seemed most intuitive but I am a bit lost.

adamltyson:
> I read the documentation but Im a bit confused about differentiating between ars and als for symmetrical atlases. I put als as i thought left to right seemed most intuitive but I am a bit lost.

In principle that shouldn't make a difference.

PolarBean:
> A bit more printing (e..g just to say what's being downloaded might be useful too).

it actually only downloads a single file as all the atlas data is distributed in a single archive. I'll add print statements elsewhere though

PolarBean:
ok actually i added one more print statement but it seems now every function has at least one print statement. I think this I've addressed everything now but feel free to add more reccomendations. I'll also update the template script with print statements so it should be included by default in the future.

PolarBean:
I ran the validation script Alessandro sent me and it returned the following error
```
Failed validations:
('catch_missing_mesh_files', "Structures with IDs [545, 1051] are in the atlas, but don't have a corresponding mesh file.")
WARNING: Could not load the Qt platform plugin "xcb" in "" even though it was found.
WARNING: This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, minimal, minimalegl, offscreen, vnc, webgl, xcb.

Aborted (core dumped)
```

I'll try to fix the xcb error but the missing structures are likely just due to those regions being too small in this atlas.

PolarBean:
image
ok after resolving the difficulties the validation script ran and showed me this volume. It all looks as expected!

alessandrofelder:
Thanks @PolarBean - I'll review next week!

PolarBean:
great! how'd you go?

brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_stereotaxic_mri_mouse.py
```diff
@@ -0,0 +1,344 @@
...
+
+# The scientific name of the species, ie; Rattus norvegicus
+SPECIES = "Mus musculus"
+
+# The URL for the data files
+ATLAS_LINK = "https://www.neuropedia.dk/resource/multimodal-3d-mouse-brain-atlas-framework-with-the-skull-derived-coordinate-system/"
+ATLAS_FILE_URL = "https://www.neuropedia.dk/wp-content/uploads/Multimodal_mouse_brain_atlas_files_v2.7z"
+# The orientation of the **original** atlas data, in BrainGlobe convention:
+# https://brainglobe.info/documentation/setting-up/image-definition.html#orientation
+ORIENTATION = "als"
```
alessandrofelder, line 86:
```suggestion
ORIENTATION = "lai"
```

alessandrofelder (commented):
Almost there.

Looks like I still get two regions without a mesh in the validation:
>('catch_missing_mesh_files', "Structures with IDs [545, 1051] are in the atlas, but don't have a corresponding mesh file.")

But my guess is those regions are so small (1051 is one pixel in the annotation, I think) that our code didn't mesh them "on purpose".

The atlas doesn't come out oriented in BrainGlobe ASR for me though, maybe [because of our `load_any` function ignoring nifti metadata](https://github.com/brainglobe/brainglobe-atlasapi/pull/370#issuecomment-2346746480). I think the original data is LAI in brainglobe terms (from looking at the raw MRI data in napari, loaded with `load_any`. 

I'll run again with my one-line suggestion to check this makes the BG version ASR and report back here.

README.md
```diff
@@ -41,6 +41,7 @@ A number of atlases are in development, but those available currently are:
 | [BlueBrain Barrel Cortex Atlas](https://doi.org/10.1162/imag_a_00209) | 10 and 25 micron | P56 | STPT | allen_mouse_bluebrain_barrels_10um, allen_mouse_bluebrain_barrels_25um
 | [UNAM Axolotl Brain Atlas](https://doi.org/10.1038/s41598-021-89357-3) | 40 micron | ~ 3 months post hatching | MRI | unam_axolotl_40um
 | [Prairie Vole Brain Atlas](https://doi.org/10.7554/eLife.87029.3.sa0) | 25 micron | Unknown | LSFM | prairie_vole_25um
+| [Gubra's stereotaxic MRI mouse brain atlas](https://doi.org/10.1007/s12021-023-09623-9) | 25 micron | 10-week-old  | MRI
```
alessandrofelder, line 44:
```suggestion
| [Gubra's stereotaxic MRI mouse brain atlas](https://doi.org/10.1007/s12021-023-09623-9) | 25 micron | 10-week-old  | MRI | perens_stereotaxic_mouse_mri_25um
```

alessandrofelder (approved):
Atlas is on GIN - thanks for your patience @PolarBean

Issue #399, opened 2024-08-28
[Feature] Template script for atlas integration


PolarBean:
**Is your feature request related to a problem? Please describe.**
I feel at the moment atlas integration is a bit complex. It seems like the atlas integration scripts can become a bit unwieldy, with unclear separation between each stage. 

**Describe the solution you'd like**
Have a simplified template script for atlas integrators to use. The basic idea would be set it up such that there are several functions that a user must fill out where the output values are clearly defined. here is a basic untested sketch of the idea which needs more work. We could as well move a lot of mesh creation logic into helper functions. Let me know your thoughts. 
```python
###Metadata
__version__ = 1 # The version of the atlas in the brainglobe_atlasapi, this is internal, if this is the first time this atlas has been added the value should be 1
ATLAS_NAME = None # The expected format is FirstAuthor_SpeciesCommonName, ie; kleven_rat
CITATION = None # DOI of the most relevant citable document
SPECIES = None # The scientific name of the species, ie; Rattus norvegicus
ATLAS_LINK = None #The URL for the data files
ORIENTATION = None #The orientation of the atlas, for more information on how to determine this click here: ........
ROOT_ID = None # The id of the highest level of the atlas. This is commonly called root or brain. Include some information on what to do if your atlas is not hierarchical
RESOLUTION = None # the resolution of your volume in microns. details on how to format this parameter for non isotropic datasets or datasets with multiple resolutions. 

def download_resources():
    """
    Download the necessary resources for the atlas.
    
    If possible, please use the Pooch library to retrieve any resources.
    """
    pass

def retrieve_template_and_reference():
    """
    Retrieve the desired template and reference as two numpy arrays.

    Returns:
        tuple: A tuple containing two numpy arrays. The first array is the template volume, and the second array is the reference volume.
    """
    template = None
    reference = None
    return template, reference

def retrieve_hemisphere_map():
    """
    Retrieve a hemisphere map for the atlas.

    If your atlas is asymmetrical, you may want to use a hemisphere map. This is an array in the same shape as your template, 
    with 0's marking the left hemisphere, and 1's marking the right.

    If your atlas is symmetrical, ignore this function.

    Returns:
        numpy.array or None: A numpy array representing the hemisphere map, or None if the atlas is symmetrical.
    """
    return None

def retrieve_structure_information():
    """
    This function should return a pandas DataFrame with information about your atlas.

    The DataFrame should be in the following format:

[...]

PR #400, opened 2024-08-28
Atlas packaging template

PolarBean:
Following on from the discussion at https://github.com/brainglobe/brainglobe-atlasapi/issues/399 I have created a functional version of the template script and updated example mouse to follow the new template.


brainglobe_atlasapi/atlas_generation/atlas_scripts/example_mouse.py
```diff
@@ -8,60 +15,103 @@
...
-    struct_tree = spacecache.get_structure_tree()  # structures tree
+    Returns:
+        numpy.array or None: A numpy array representing the hemisphere map, or None if the atlas is symmetrical.
+    """
+    return None
+
+
+def retrieve_structure_information():
+    """
+    Retrieve the structures tree and meshes.
```
alessandrofelder, line 67:
```suggestion
    Retrieve the structures tree and meshes for the Allen mouse brain atlas.
```

brainglobe_atlasapi/atlas_generation/atlas_scripts/template_script.py
```diff
@@ -0,0 +1,111 @@
+from brainglobe_atlasapi.atlas_generation.wrapup import wrapup_atlas_from_data
+
+###Metadata
+__version__ = 1  # The version of the atlas in the brainglobe_atlasapi, this is internal, if this is the first time this atlas has been added the value should be 1
+ATLAS_NAME = None  # The expected format is FirstAuthor_SpeciesCommonName, ie; kleven_rat
+CITATION = None  # DOI of the most relevant citable document
+SPECIES = None  # The scientific name of the species, ie; Rattus norvegicus
+ATLAS_LINK = None  # The URL for the data files
+ORIENTATION = None  # The orientation of the atlas, for more information on how to determine this click here: ........
```
alessandrofelder, line 9:
```suggestion
ORIENTATION = "asr"  # The orientation of the **original** atlas data, in BrainGlobe convention: https://brainglobe.info/documentation/setting-up/image-definition.html#orientation
```

brainglobe_atlasapi/atlas_generation/atlas_scripts/template_script.py
```diff
@@ -0,0 +1,111 @@
...
+ROOT_ID = None  # The id of the highest level of the atlas. This is commonly called root or brain. Include some information on what to do if your atlas is not hierarchical
```
alessandrofelder, line 10: > Include some information on what to do if your atlas is not hierarchical

Not sure I understand this: is this information we should include to help people running this script?

brainglobe_atlasapi/atlas_generation/atlas_scripts/example_mouse.py
```diff
@@ -8,60 +15,103 @@
...
+
+
+def retrieve_hemisphere_map():
+    """
+    Retrieve a hemisphere map for the atlas.
+
+    If your atlas is asymmetrical, you may want to use a hemisphere map. This is an array in the same shape as your template,
+    with 0's marking the left hemisphere, and 1's marking the right.
+
+    If your atlas is symmetrical, ignore this function.
```
alessandrofelder, line 57:
```suggestion
The Allen atlas is symmetrical, so we can just return `None` in this function.
```

brainglobe_atlasapi/atlas_generation/atlas_scripts/example_mouse.py
```diff
@@ -1,4 +1,11 @@
-__version__ = "2"
+__version__ = 1  # The version of the atlas in the brainglobe_atlasapi, this is internal, if this is the first time this atlas has been added the value should be 1
+ATLAS_NAME = "example_mouse"  # The expected format is FirstAuthor_SpeciesCommonName, i.e., kleven_rat
+CITATION = "Wang et al 2020, https://doi.org/10.1016/j.cell.2020.04.007"  # DOI of the most relevant citable document
+SPECIES = "Mus musculus"  # The scientific name of the species, i.e., Rattus norvegicus
+ATLAS_LINK = "http://www.brain-map.org"  # The URL for the data files
+ORIENTATION = "asr"  # The orientation of the atlas
+ROOT_ID = 997  # The id of the highest level of the atlas. This is commonly called root or brain.
+RESOLUTION = 100  # The resolution of your volume in microns.
```
alessandrofelder, line 8:
```suggestion
# a working example atlas-packaging script, which makes a simplified version of the Allen Mouse Brain Atlas, at 100um resolution. See `template_script.py` for a starting point to package your own atlas.
__version__ = 0  # This will make the example mouse version 1.0 (zero is the minor version)
ATLAS_NAME = "example_mouse" 
CITATION = "Wang et al 2020, https://doi.org/10.1016/j.cell.2020.04.007"  # DOI of the most relevant citable document
SPECIES = "Mus musculus"  # The scientific name of the species,
ATLAS_LINK = "http://www.brain-map.org"  # The URL for the data files
ORIENTATION = "asr"  # The orientation of the atlas
ROOT_ID = 997  # The id of the highest level of the atlas. This is commonly called root or brain.
RESOLUTION = 100  # The resolution of your volume in microns.
```
Avoid duplication of information with the template script, and make the difference between the scripts clearer?

brainglobe_atlasapi/atlas_generation/atlas_scripts/template_script.py
```diff
@@ -0,0 +1,111 @@
+from brainglobe_atlasapi.atlas_generation.wrapup import wrapup_atlas_from_data
+
```
alessandrofelder, line 2:
```suggestion
# Copy-paste this script into a new file and fill in the functions to package your own atlas.
```

brainglobe_atlasapi/atlas_generation/atlas_scripts/template_script.py
```diff
@@ -0,0 +1,111 @@
+from brainglobe_atlasapi.atlas_generation.wrapup import wrapup_atlas_from_data
```
alessandrofelder, line 1:
```suggestion
from brainglobe_atlasapi.atlas_generation.wrapup import wrapup_atlas_from_data
from pathlib import Path
```
Make linter happy.

brainglobe_atlasapi/atlas_generation/atlas_scripts/template_script.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    meshes_dict = {}
+    return meshes_dict
+
+
+### If the code above this line has been filled correctly, nothing needs to be edited below (unless variables need to be passed between the functions).
+bg_root_dir = Path.home() / "brainglobe_workingdir" / ATLAS_NAME
+bg_root_dir.mkdir(exist_ok=True)
+download_resources()
+template_volume, reference_volume = retrieve_template_and_reference()
```
alessandrofelder, line 88:
```suggestion
template_volume, annotated_volume = retrieve_reference_and_annotation()
```

brainglobe_atlasapi/atlas_generation/atlas_scripts/template_script.py
```diff
@@ -0,0 +1,111 @@
...
+def download_resources():
+    """
+    Download the necessary resources for the atlas.
+
+    If possible, please use the Pooch library to retrieve any resources.
+    """
+    pass
+
+
+def retrieve_template_and_reference():
```
alessandrofelder, line 23:
```suggestion
def retrieve_reference_and_annotation():
```

brainglobe_atlasapi/atlas_generation/atlas_scripts/template_script.py
```diff
@@ -0,0 +1,111 @@
...
+    ╭─────────────┬───────────────────────────────────┬─────────┬───────────────────────┬───────────────────╮
+    | id          | name                              | acronym | structure_id_path     | rgb_triplet       |
+    |             |                                   |         |                       |                   |
+    ├─────────────┼───────────────────────────────────┼─────────┼───────────────────────┼───────────────────┤
+    | 997         | root                              | root    | []                    | [255, 255, 255]   |
+    ├─────────────┼───────────────────────────────────┼─────────┼───────────────────────┼───────────────────┤
+    | 8           | Basic cell groups and regions     | grey    | [997]                 | [191, 218, 227]   |
+    ├─────────────┼───────────────────────────────────┼─────────┼───────────────────────┼───────────────────┤
+    | 567         | Cerebrum                          | CH      | [997, 8]              | [176, 240, 255]   |
+    ╰─────────────┴───────────────────────────────────┴─────────┴───────────────────────┴───────────────────╯
```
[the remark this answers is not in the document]
alessandrofelder: But we need to make the linter happy, so maybe we can squish it a bit horizontally? :crossed_fingers:

alessandrofelder (changes requested):
Thanks a lot for this @PolarBean - looking good.

I've made many comments, but most of them are small. I've tried to reduce duplicate info by making the example script doc-strings cater to people that like concrete examples, and make the template script doc-strings more instructions about what one needs to do.

I still kind of think we could do with just one and not the other, but happy to keep both for now. We can think more about this as we refactor the API and head towards version 2.

As always, I am open to counter-arguments to my suggestions :slightly_smiling_face: 

(I've made some suggestions to make the linter happ*ier* with the template script, but it's unlikely to have covered all of it. That also needs to be addressed before we merge this.

Some questions I have:
* should we explain somewhere how to include additional references or is that considered out-of-scope here?
* should we point to mesh helper functions somewhere?

adamltyson:
> should we explain somewhere how to include additional references or is that considered out-of-scope here?

Should we have an explicit section like:
```python
# **** OPTIONAL *****
additional reference stuff
```

brainglobe_atlasapi/atlas_generation/atlas_scripts/template_script.py
```diff
@@ -0,0 +1,111 @@
...
+    This function should return a dictionary of ids and corresponding paths to mesh files.
+    Some atlases are packaged with mesh files, in these cases we should use these files.
+    Then this function should download those meshes. In other cases we need to construct
+    the meshes ourselves. For this we have helper functions to achieve this.
+    """
+    meshes_dict = {}
+    return meshes_dict
+
+
+### If the code above this line has been filled correctly, nothing needs to be edited below (unless variables need to be passed between the functions).
```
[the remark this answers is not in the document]
alessandrofelder: Belated thought. This (and its equivalent in the example script) should be inside `if __name__ == "__main__" guards` to avoid atlas creation code being executed when this module is imported.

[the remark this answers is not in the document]
PolarBean: sure thing, i hadn't considered that

PolarBean:
> > should we explain somewhere how to include additional references or is that considered out-of-scope here?
> 
> Should we have an explicit section like:
> 
> ```python
> # **** OPTIONAL *****
> additional reference stuff
> ```

this seems like a good solution but i cant see any instances of multiple an atlas that has multiple citable documents (if this is what you're referring to). How should this be implemented?

brainglobe_atlasapi/atlas_generation/atlas_scripts/example_mouse.py
```diff
@@ -8,60 +15,103 @@
...
-    download_dir_path = working_dir / "downloading_path"
+def retrieve_template_and_reference():
+    """
+    Retrieve the desired template and reference as two numpy arrays.
+
+    Returns:
+        tuple: A tuple containing two numpy arrays. The first array is the reference volume, and the second array is the annotated volume.
+    """
+    # Create temporary download directory
+    download_dir_path = Path.cwd() / "downloading_path"
```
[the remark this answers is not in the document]
PolarBean: good suggestion, I am now using it to validate the hashes of the files downloaded via the API. 
https://github.com/PolarBean/brainglobe-atlasapi/blob/atlas_packaging_template/brainglobe_atlasapi/atlas_generation/atlas_scripts/example_mouse.py#L53-L73

brainglobe_atlasapi/atlas_generation/atlas_scripts/template_script.py
PolarBean: As in people will use this script to package their own atlases. We should explain how to handle particular edge cases they may run into. for example what if the atlas is not hierarchical? Do we then just say treat everything having a  structure_id_path of brain, region? or what if their atlas doesnt have RGB colours assigned to each region, is it then acceptable for them to randomly generate colours? I think the template should either explain these cases or point to a resource that does.

PolarBean:
@alessandrofelder I've addressed your comments and am ready for this to be re-reviewed once you have the time :)

adamltyson:
> this seems like a good solution but i cant see any instances of multiple an atlas that has multiple citable documents (if this is what you're referring to). How should this be implemented?

Here we mean multiple reference images (i.e. templates, we should rename this!). I think the AZBA has 2 reference images.

brainglobe_atlasapi/atlas_generation/atlas_scripts/example_mouse.py
```diff
@@ -10,58 +9,135 @@
...
+ORIENTATION = "asr"  # The orientation of the atlas
+ROOT_ID = 997  # The id of the highest level of the atlas.
+RESOLUTION = 100  # The resolution of your volume in microns.
+
+BG_ROOT_DIR = Path.home() / "brainglobe_workingdir" / ATLAS_NAME
+
+
+def download_resources():
+    """
+    Download the necessary resources for the atlas.
```
alessandrofelder, line 29:
```suggestion
    Download the necessary resources for the atlas. Here we don't, because we can out-source this to the Allen SDK in later functions.
```

brainglobe_atlasapi/atlas_generation/atlas_scripts/example_mouse.py
```diff
@@ -10,58 +9,135 @@
...
-    print("Download completed...")
+    reference_volume, _ = spacecache.get_template_volume()
+    annotation_volume, _ = spacecache.get_annotation_volume()
+    expected_reference_hash = (
+        "6c24cae773a5cf256586b0384af0ac93ad68564d211c9bdcff4bee9acf07786a"
+    )
+    expected_annotation_hash = (
+        "451e6a82f531d3db4b58056d024d3e2311703c2adc15cefa75e0268e7f0e69a4"
+    )
+    reference_hash = pooch.file_hash(
```
alessandrofelder, line 63: I worry that this is
a) not the way we typically use `pooch` and therefore confusing
b) more importantly, might look scary to new contributors coming more from the biology side

I'm tempted to remove any `pooch`-related things, and mention it in the template script only (which we already do in this PR)?

alessandrofelder (changes requested):
Thanks a lot @PolarBean - looking good!
Two more tiny comments, then we can merge this.

brainglobe_atlasapi/atlas_generation/atlas_scripts/example_mouse.py
PolarBean: I see your point, the only thing is that pooch here actually serves a purpose, to ensure that the files from the Allen don't change without notifying us. In this instance we would want pooch to confirm the hash of the file. I understand that it's non standard but I think that just comes with using the Allen API in the example. 

An alternative would be to not use the allen API at all, instead just use the files from the [allen informatics archive](download.alleninstitute.org/informatics-archive/current-release/mouse_ccf/). The advantage to this is the example would become more of a standard atlas integration allowing us to use pooch, etc, instead of so much allen specific logic.

alessandrofelder: >An alternative would be to not use the allen API at all, instead just use the files from the [allen informatics archive](https://github.com/brainglobe/brainglobe-atlasapi/pull/download.alleninstitute.org/informatics-archive/current-release/mouse_ccf/).

I think this would be great if you have the patience to implement this? :grin: Would this be possible/easy on the `OntologiesAPI` side too?

The alternative is that we make a new issue to at some point move to a more standard atlas (e.g. we have recently generated a bird template for which we will host the data ourselves, and which has ITK snap annotations - our standard - so we could use that when the preprint is published.)

alessandrofelder:
I am tempted to merge this at some point this week, and open a new issue for moving it away from AllenSDK - any objections @PolarBean @adamltyson ?

adamltyson:
No argument from me.

PolarBean:
Works for me!

brainglobe_atlasapi/atlas_generation/atlas_scripts/example_mouse.py
alessandrofelder: We can tackle this separately with #416 .

PR #405, opened 2024-09-05
don't rerun tests on README updates

PolarBean:
currently the tests rerun on all pushes. Even if that push was just to update the README.
very minor change, it just bugged me seeing the tests always rerun on readme changes :)

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -6,6 +6,8 @@ on:
       - "main"
     tags:
       - "v**"
+    paths-ignore:
+      - '**/README.md'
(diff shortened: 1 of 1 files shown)


adamltyson:
I would err on the side of keeping the current behaviour. Currently our tests don't include any text linting, but they may do in the future. We also have 100+ repos to maintain, so I like to keep them as similar as possible :)

Happy to be overruled though. Any thoughts @brainglobe/active-devs?

PolarBean:
no problem. Happy to close it!

adamltyson:
Thanks @PolarBean. I will close this now, but it may be an idea we come back to.

Issue #319, opened 2024-07-05
[Feature] Add the Australian Mouse Brain Mapping Consortium Atlas


PolarBean:
The Australian Mouse Brain Mapping Consortium released a fantastic 15um resolution population average MRI atlas.
https://imaging.org.au/AMBMC/
The template is great but the labels are a bit limited, not covering the whole brain. It is in Waxholm Mouse Space 

It could be a good resource for those working with MRI. It is available under a Creative Commons Attribution 3.0 Australia Licence so is open source just requiring attribution. 

The recommended citation is the following conference paper 

Janke AL, Ullmann J, Kurniawan N, Paxinos G, Keller M, Yang Z, Richards K, Egan G, Petrou S, Galloway G, Reutens D. (2012) 15μm average mouse models in Waxholm space from 16.4T 30μm images. In 20th Annual ISMRM Scientific Meeting and Exhibition, Melbourne, Australia.

However I cant find this online. It is also described in the following papers that I can see. 

[A segmentation protocol and MRI atlas of the C57BL/6J mouse neocortex](https://www.sciencedirect.com/science/article/pii/S1053811913003352#f0015)
[An ontologically consistent MRI-based atlas of the mouse diencephalon](https://www.sciencedirect.com/science/article/pii/S1053811917304561)
[Segmentation of the C57BL/6J mouse cerebellum in magnetic resonance images](https://www.sciencedirect.com/science/article/pii/S1053811912005496)
[Segmentation of the mouse hippocampal formation in magnetic resonance images](https://www.sciencedirect.com/science/article/pii/S1053811911006434#f0005)
[An MRI atlas of the mouse basal ganglia](https://link.springer.com/article/10.1007/s00429-013-0572-0)

adamltyson:
This looks nice. I'm not super familiar with the Waxholm mouse space. Are there other atlases in this space with better annotations?

BTW I think there's just the abstract for ISMRM meetings, and it's [here](https://cds.ismrm.org/protected/12MProceedings/SearchResults.php?q=average+mouse+models+in+Waxholm).

PolarBean:
There are other annotations such as https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2930145/ which is shared https://www.nitrc.org/projects/incfwhsmouse but I don't think they are better. 

I haven't checked to see whether these two templates align with each other but I would hope so. 

I don't know how much utility these have but I will still integrate them into the api soon.

[...]

PR #409, opened 2024-09-12
Australian mouse brain

PolarBean:
This is to add the australian mouse brain atlas, resolving this issue. 
https://github.com/brainglobe/brainglobe-atlasapi/issues/319
The benefits of this atlas are that it is a 15µm resolution MRI template. Which I believe is currently the highest resolution MRI template by quite a distance (Though they used a novel method to upsample from 30µm described here: https://cds.ismrm.org/protected/12MProceedings/PDFfiles/1077.pdf.)
Also these annotations are at least partially based on the Paxinos ontology and Paxinos is an author which may be useful for comparative purposes as the Paxinos atlas is widely used for 2D atlasing. 
The challenges when adding this atlas that I faced were the following:

- The annotations were stored across multiple files with multiple corresponding label metadata
- The IDs were overlapping between files which required renumbering to resolve
- The acronyms for each ID were given but for the full name I had to search for tables in the corresponding paper for that region.
- The acronym to full name tables were formatted differently in different journals, in one case it was a png. This is why I had to include the JSON that I constructed in the atlas generation script here.
- In some limited cases it wasnt entirely clear which full name an acronym corresponded to which required me making an educated guess. Some were obvious, others less so.

Nonetheless, hopefully this is a useful addition to the API.


adamltyson:
Thanks @PolarBean! Bear with us, it might be a few days until we can review this.

PolarBean:
image
Fixed some small issues discovered after running the validation script and all appears to run now.

alessandrofelder:
Thanks @PolarBean - I will run it locally this week, and upload it to GIN assuming I get the validation to pass to :)

brainglobe_atlasapi/atlas_generation/atlas_scripts/australian_mouse_brain.py
```diff
@@ -0,0 +1,900 @@
...
+}
+REFERENCE_URL = "https://osf.io/download/g8p6a"
+REGION_IDS = {
+    "basalganglia": 1001,
+    "cerebellum": 1002,
+    "cortex": 1003,
+    "hippocampus": 1004,
+    "diencephalon": 1005,
+}
+PARALLEL = False
```
alessandrofelder, line 51: I think we should remove the parallelisation here and where it's used below (see #428)

alessandrofelder (commented):
Thanks a lot for this @PolarBean - so much effort!

I've run this locally, and it's largely looking very good. I think there's two things we need to sort out with the annotations (at least for the result I get):
* some spurious pixels (e.g. on coronal slice 9238) outside the brain that claim to be part of CA1-Pyramidal oriens layer!
[image](https://github.com/user-attachments/assets/8b1dab8d-5e1b-435e-bd79-ceda7e86574f)
* it looks like the annotations file does not have pixels labelled with [35, 36, 37, 38, 39, 40, 41, 42, 43, 44, 45, 62, 64, 256, 257, 258, 259, 260, 261, 262] but somehow these make their way into the metadata, leading to the validation complaining about these regions not having meshes associated with them. Is this expected?

I've also made some tiny nitpicky comments about the code.

PolarBean:
>     * in the annotation image, there are some spurious pixels (e.g. on coronal slice 9238) outside the brain that claim to be part of CA1-Pyramidal oriens layer!
>       [image](https://github.com/user-attachments/assets/8b1dab8d-5e1b-435e-bd79-ceda7e86574f)
These are in the atlas file shared by the creator so I would say leave them (even though you are correct they are a mistake). The only other alternative I can think of is brainglobe having an automated filter for out of brain pixels. But then we are getting into the territory of creating a brainglobe version of each atlas which is unique to brainglobe. I'm not fussed which way we choose but we should be consistent. 

>     * it looks like the annotations file does not have pixels labelled with [35, 36, 37, 38, 39, 40, 41, 42, 43, 44, 45, 62, 64, 256, 257, 258, 259, 260, 261, 262] but somehow these make their way into the metadata, leading to the validation complaining about these regions not having meshes associated with them. Is this expected?
> 
Yeah this is expected :) For instance you have layers 1, 2, 3 of some structure but only layers 1 and 3 are actually delineated. 2 can still exist in the metadata as it should exist just hasn't been delineated yet. If we want to create a rule that these should be filtered out then we would run into the same issue described above. 


Ill go through the code comments now!

brainglobe_atlasapi/atlas_generation/atlas_scripts/australian_mouse_brain.py
```diff
@@ -0,0 +1,900 @@
...
+
+def download_resources():
+    DOWNLOAD_DIR_PATH.mkdir(exist_ok=True)
+    ## Download atlas_file
+    # utils.check_internet_connection()
+    destination_path = DOWNLOAD_DIR_PATH / "template.nii.tar.gz"
+    if not os.path.isfile(destination_path):
+        response = requests.get(REFERENCE_URL, stream=True)
+        with open(destination_path, "wb") as f:
+            for chunk in response.iter_content(chunk_size=8192):
```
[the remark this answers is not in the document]
PolarBean: 8192 bits is 1 kibibyte. but yeah it's arbitrary. I can add a comment on this

brainglobe_atlasapi/atlas_generation/atlas_scripts/australian_mouse_brain.py
PolarBean: removed

adamltyson:
> These are in the atlas file shared by the creator so I would say leave them (even though you are correct they are a mistake). The only other alternative I can think of is brainglobe having an automated filter for out of brain pixels. But then we are getting into the territory of creating a brainglobe version of each atlas which is unique to brainglobe. I'm not fussed which way we choose but we should be consistent.

Previously we decided (and should document) that in these cases we would release a version without “editing” anything, and then a later version with the corrections. Not sure we’ve done this yet, but if we do we should make sure it’s clearly documented.

adamltyson:
> Yeah this is expected :) For instance you have layers 1, 2, 3 of some structure but only layers 1 and 3 are actually delineated. 2 can still exist in the metadata as it should exist just hasn't been delineated yet. If we want to create a rule that these should be filtered out then we would run into the same issue described above.

Happy to be outvoted, but I think in this example, 2 shouldn’t be in the metadata, as it doesn’t actually exist in the atlas. This would be consistent with other BG atlases anyway.

PolarBean:
I disagree, as the ontology itself is valuable. If someone creates a highly detailed ontology, even if it has not yet been fully delineated, I would like to access that through brainglobe.

adamltyson:
I can see why it would be useful, but to add that information, we would need to change the API. Currently third party tools rely on e.g. a 1-2-1 mapping between region in the ontology and a mesh.

I've raised an issue to track this (#439)

PolarBean:
Great! For now then I'll remove the regions listed as not existing in the volume :)

alessandrofelder:
> For now then I'll remove the regions listed as not existing in the volume :)

Thanks!!! Let me know when you've done so - then we can merge this :tada: and I'll upload to GIN.

alessandrofelder:
Hey @PolarBean 
Trying to get back on top of all PRs on BrainGlobe after the winter break - am I right in thinking that the next step here is for you to remove the regions that are not yet delineated? Sorry if I missed something and let me know when this is ready for review again :)

PolarBean:
I've removed those regions and it should be ready.

PolarBean:
Just checking to see whether theres anything I need to do on this!

alessandrofelder:
Thanks for checking in @PolarBean - I ran it on my macbook on Friday and I got an `OverflowError`. Will try again this afternoon on a different machine to help narrow down the problem and report back 🤔

PolarBean:
Hmm how strange. I'll let you know if I run into anything like that

alessandrofelder (approved):
This seems to work and generate a valid atlas (including for the validation functions in #453 ) on my Ubuntu so I'm happy with this. Visual inspection also looks reasonable.

alessandrofelder:
I've uploaded to GIN and checked I can instantiate a `BrainGlobeAtlas` object from the downloaded GIN data. @PolarBean would you have time to add this atlas to the docs here and on the website repo?

PolarBean:
Absolutely! I'll get on it tomorrow.

PR #417, opened 2024-10-29
Make list atlas function more robust

adamltyson:
The existing function returned garbage if there was a "_v" anywhere in the atlas name (e.g. "prairie_vole"), and also didn't produce anything informative if the atlas couldn't be found. This PR makes this function more robust and adds error handling.

### DIFF
--- a/brainglobe_atlasapi/list_atlases.py
+++ b/brainglobe_atlasapi/list_atlases.py
@@ -2,6 +2,8 @@
     Some functionality to list all available and downloaded brainglobe atlases
 """
 
+import re
+
 from rich import print as rprint
 from rich.panel import Panel
 from rich.table import Table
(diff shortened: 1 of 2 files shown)


brainglobe_atlasapi/list_atlases.py
```diff
@@ -43,11 +45,15 @@ def get_local_atlas_version(atlas_name):
...
-    ][0]
+    try:
+        return [
+            re.search(r"_v(\d+\.\d+)$", f.name).group(1)
+            for f in brainglobe_dir.glob(f"*{atlas_name}*")
+            if f.is_dir() and re.search(r"_v(\d+\.\d+)$", f.name)
+        ][0]
+    except IndexError:
+        print(f"No atlas found with the name: {atlas_name}")
+        return None
```
alessandrofelder, line 56: Might be good to add a very simple test that checks this error is raised if the regex is not matched?

alessandrofelder (approved):
Suggestion to add a test, but otherwise happy.

brainglobe_atlasapi/list_atlases.py
adamltyson: Good shout. I was so far down the rabbit hole I forgot about tests!

alessandrofelder (approved):

PR #422, opened 2024-10-30
Add api name of atlases to readme

PolarBean:
One difficulty I run into is I read about an atlas in the readme, and then struggle to find it in the API as the name is not always clear from the information provided in the table. Here I suggest we add the full API names to the table.


adamltyson:
This is great, I think it should have been added a while ago. It takes up a lot of space, but the atlases are the most important aspect of the API, so I think it's worth it.

adamltyson (approved):

PR #430, opened 2024-11-15
Fail gracefully if cache directory is read only

IgorTatarnikov:

## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**
If a user doesn't have write permission to the `.brainglobe` directory, instantiating an atlas will cause a crash unless `check_latest=False` is set.

**What does this PR do?**
Gracefully handles permission denied errors when caching the `latest_versions.conf` file. The user will be informed of the error without crashes.

## References
https://github.com/brainglobe/brainglobe-heatmap/issues/65
https://github.com/brainglobe/brainrender/issues/395

Tested locally by changing the permissions for the `last_versions.conf` file.

New test added to check error is printed once, and `utils.conf_from_url` runs even if `.brainglobe` has read only permissions.

## Is this a breaking change?
No.

No.

## Checklist:

### DIFF
--- a/brainglobe_atlasapi/utils.py
+++ b/brainglobe_atlasapi/utils.py
@@ -291,7 +291,7 @@ def get_download_size(url: str) -> int:
 
 
 def conf_from_url(url) -> configparser.ConfigParser:
-    """Read conf file from an URL. And cache a copy in the brainglobe dir.
+    """Read conf file from a URL. And cache a copy in the brainglobe dir.
     Parameters
     ----------
     url : str
@@ -305,14 +305,17 @@ def conf_from_url(url) -> configparser.ConfigParser:
     text = requests.get(url).text
     config_obj = configparser.ConfigParser()
     config_obj.read_string(text)
-    cache_path = config.get_brainglobe_dir() / "last_versions.conf"
+    cache_path: Path = config.get_brainglobe_dir() / "last_versions.conf"
 
-    if not cache_path.parent.exists():
-        cache_path.parent.mkdir(parents=True, exist_ok=True)
+    try:
+        if not cache_path.parent.exists():
+            cache_path.parent.mkdir(parents=True, exist_ok=True)
 
-    # Cache the available atlases
-    with open(cache_path, "w") as f_out:
-        config_obj.write(f_out)
+        # Cache the available atlases
+        with open(cache_path, "w") as f_out:
+            config_obj.write(f_out)
+    except OSError as e:
+        print(f"Could not update the latest atlas versions cache: {e}")
 
     return config_obj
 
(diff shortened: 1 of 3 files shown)


brainglobe_atlasapi/utils.py
```diff
@@ -305,14 +305,17 @@ def conf_from_url(url) -> configparser.ConfigParser:
...
-        config_obj.write(f_out)
+    try:
+        if not cache_path.parent.exists():
+            cache_path.parent.mkdir(parents=True, exist_ok=True)
+
+        # Cache the available atlases
+        with open(cache_path, "w") as f_out:
+            config_obj.write(f_out)
+    except OSError as e:
+        print(f"Could not update the latest atlas versions cache: {e}")
```
alessandrofelder, line 318: Would it be better if this used `logging.warning` or `logging.info`?

alessandrofelder (approved):
I am happy with the changes - made a very minor comment.
To check my understanding: this means that if `latest_versions.conf` is read-only, users won't be able to update atlases (and will need e.g. a sysadmin to do it for them)?

brainglobe_atlasapi/utils.py
IgorTatarnikov: I wasn't sure, the rest of the file uses `print` statements to inform user of error cases so I didn't want to break from what's currently being used. Does `brainglobe-atlasapi` keep a log on disk somewhere?

alessandrofelder: It looks like earlier in the `utils.py` file it sets the logging level for `urllib3` but if `print` is used elsewhere, maybe let's stick to `print` here and open an issue to make logging consistent... (`atlasgen` uses `loguru`)

IgorTatarnikov:
I think so, if the `.brainglobe` directory is read only then the user can't update the atlases anyway (as far as I understand).

- If the internet is available, the directory is read only, and `last_versions.conf` file doesn't exists: the user gets the printed warning, atlases are loaded.

- If the internet is available, the directory is read only, and `last_versions.conf` exists: the user gets the printed warning, atlases are loaded.

- If there is no internet, the directory is read only, and `last_versions.conf` exists:  no printed warning, atlases are loaded.

- If there is no internet, the directory is read only, and `last_versions.conf` doesn't exist: no printed warning, atlases are loaded.

Thinking more about it, is it worth getting rid of the printed warning?

IgorTatarnikov:
Currently, they only get a warning that their directory is read only, nothing to do with being able to get the latest version of an atlas. If they're connected to the internet the atlas version check is done by comparing to the data fetched from GIN not the cached `last_versions.conf`.

Issue #425, opened 2024-11-06
[Feature] DevCCFv1 atlas


carlocastoldi:
related pubblication: https://www.nature.com/articles/s41467-024-53254-w
data: https://pennstateoffice365-my.sharepoint.com/:f:/g/personal/yuk17_psu_edu/EmCllFDonwtLvDD0xgWd7QYBuzVVvnSv4oKpUy7F9bx75Q?e=RxAmJa

website: https://kimlab.io/brain-map/DevCCF/

adamltyson:
Thanks for raising this @carlocastoldi. I'm going to close this though as it's a duplocate of https://github.com/brainglobe/brainglobe-atlasapi/issues/225

adamltyson:
If you need this atlas, and you'd like to contribute it to BrainGlobe, please let us know. We'd be happy to help.

carlocastoldi:
Oh thank you!
I would love to contribute to BrainGlobe like i've done in the past. However it might not be something I can work on in the near future.

If i start working on it, I'll let you know!

Issue #429, opened 2024-11-11
[Feature] Speed up packaging for atlases, e.g. by parallelising mesh generation (robustly!)


alessandrofelder:
**Is your feature request related to a problem? Please describe.**
Atlas packaging is slow for atlases with many regions.

**Describe the solution you'd like**
A longer term solution to #428 that doesn't compromise on performance - with good tests.

pranav33317:
Sir now that #428 is complete , how should I re-introduce parallel mesh creation for atlases ? Could you please clear up the tasks involved here, it would really help here?
Thanks

IgorTatarnikov:
They would, it seems I tested visualising the outputs after running it without extracting subvolumes 😅. This solution also wasn't scaling when I tried running `kim_dev_mouse.py`. The memory usage still spikes to 40/50+ GB when creating meshes for P04 onwards.

Issue #428, opened 2024-11-11
[BUG] remove unreliable parallel running of mesh creation


alessandrofelder:
**Describe the bug**
We know running mesh creation in parallel is unreliable, so we should remove it from any atlas-packaging scripts. In the first instance. This issue will be address when the results of [a search for PARALLEL in this repo](https://github.com/search?q=repo%3Abrainglobe%2Fbrainglobe-atlasapi%20PARALLEL&type=code) are empty

pranav33317:
I have submitted PR : https://github.com/brainglobe/brainglobe-atlasapi/pull/546  for this please check

alessandrofelder:
Closed by #546

Issue #141, opened 2023-02-23
BrainGlobe Atlas API Version 2


adamltyson:
This is essentially a reply to https://github.com/brainglobe/bg-atlasapi/issues/96, but I'm starting a new issue to track this idea. Sorry for the long post, but interested in your ideas, @brainglobe/maintainers @brainglobe/swc-neuroinformatics @brainglobe/czi_eoss5. 

After 2.5 years, and based on conversations with various users and atlas creators, I think it's time for `bg-atlasapi` version 2. Version 1 works very well for "classical" anatomical atlases (i.e. one reference image, and one annotation), but it doesn't cater well (or at all) for:
* Atlases with multiple reference images (e.g. `kim_dev_mouse` and `mpin_zfish`)
* Atlases that exist at multiple resolutions (the atlas API doesn't link these in any way)
* Atlases with the same annotations in different coordinate spaces (e.g. `allen_mouse` and `perens_lsfm_mouse`)
* Atlases that aren't represented by brain regions (e.g cell atlases)
* Atlases (or coordinate spaces) that incorporate other data (e.g. tracing, gene expression)
* Atlases being updated (as discussed above)

The atlas generation process also needs streamlining.

# My idea for V2:

## Move away from the monolithic atlas structure

The atlas could be defined by a config file, specifying atlas "elements" (an element being reference image, set of meshes etc):

```yaml
 name: "allen_mouse"
 atlas_link: "http://www.brain-map.org"
 version: 2.0
...
...
reference_images: 
 # could be >1
  STP:  some_url
annotation_images: 
 # could be >1
  CCFv3:  some_url
structures: some_url
meshes: some_url
```

When the atlas was downloaded, the atlas API would check to see which of these existing files had been downloaded, and then only download those required. The idea is that there would be a lot of overlap between atlases (same meshes at different resolutions, same reference image for multiple annotations etc.), and this would reduce download times and save disk space.

This would also allow data to be stored somewhere other than GIN. I'm not sure whether we want to do this, but it may be necessary for e.g. larger atlases (see below).

## Improve versioning

Essentially as per https://github.com/brainglobe/bg-atlasapi/issues/96. We could version the elements individually, and a versioned atlas could specify these, e.g.:

```yaml
 name: "allen_mouse"
 atlas_link: "http://www.brain-map.org"
 version: 1.0
...
...
reference_images: 
  STP:  some_url@v1.0.0
annotation_images: 
  CCFv3:  some_url@1.0.0
[...]

PR #441, opened 2024-12-16
Add all Kim's DevCCFv1 atlases

carlocastoldi:
## Description
With this draft PR i ask you your opinion and help to first:
* allow the script to download the resources from the official [OneDrive page](https://pennstateoffice365-my.sharepoint.com/:f:/g/personal/yuk17_psu_edu/EmCllFDonwtLvDD0xgWd7QYBuzVVvnSv4oKpUy7F9bx75Q?e=RxAmJa)
* have a general review of the script's code quality before adapting it to generate the other combination of atlases/references
* choose which atlases are worth generating and whether they should update any previous atlas

Lastly, I have tested this script on two setups: on one computer _without_ parallelization and the other _with_. The run with parallelization didn't manage the memory properly and run OOM on linux, effectively making the generation of the atlas much slower. Does BrainGlobe offer any help in the management of the parallelization or should I do it myself?

P.S. if needed, I can provide the resulting atlas from this script

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**
To perfection the support to Kim's  developing mouse brain atlases

**What does this PR do?**
It adds a script for generating [Kim's DevCCFv1](https://kimlab.io/brain-map/DevCCF/) mouse atlas. For now, it generates only the P14 altas with LSFM@20µm as reference, ignoring the other ages (E11.5, E13.5, E15.5, E18.5, P4 and P56) and references (they are all MRI@50 µm)

## References
#425

The resulting atlas from the script was used within [ABBA](https://go.epfl.ch/abba)

## Is this a breaking change?

No.

It does, but it is not included yet in this PR

## Checklist:


alessandrofelder:
Hey @carlocastoldi 

Thanks a lot!

> Does BrainGlobe offer any help in the management of the parallelization or should I do it myself?

We are aware that past parallelisation efforts are not robust https://github.com/brainglobe/brainglobe-atlasapi/issues/428, so if you would like to contribute a better way, even in just for one atlas, that might help us address https://github.com/brainglobe/brainglobe-atlasapi/issues/429 - so that would be great

> choose which atlases are worth generating and whether they should update any previous atlas

I am happy to give my thoughts on code quality and how to download from OneDrive but can I first check (Pardon my naivety!): what is the difference here to the [existing Kim Developmental atlas](https://github.com/brainglobe/brainglobe-atlasapi/blob/main/brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_developmental_ccf_mouse.py) that is already packaged? Could this be an extension of that existing script?

carlocastoldi:
> We are aware that past parallelisation efforts are not robust https://github.com/brainglobe/brainglobe-atlasapi/issues/428, so if you would like to contribute a better way, even in just for one atlas, that might help us address https://github.com/brainglobe/brainglobe-atlasapi/issues/429 - so that would be great

Oh okay, i see that! It's something I may perhaps look into, but for now it can't be my priority...

> can I first check (Pardon my naivety!): what is the difference here to the [existing Kim Developmental atlas](https://github.com/brainglobe/brainglobe-atlasapi/blob/main/brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_developmental_ccf_mouse.py) that is already packaged? Could this be an extension of that existing script?

Oh I tried to understand if the existing "developmental" atlas (it's not really developmental because it's only for P56, when mice are already adult. It's the same age at which Allen developed CCFv3) had any publication associated with it, but as far as i understood, [that's only a v0.0.1](https://data.mendeley.com/datasets/2svx788ddf/1) of the work they have recently published (v1). I am not 100% sure about it, tho, as i can't find any description/publication about that dataset.
Surely, however, it is not compatible with their v1 as they differ in resolution, coordinate system, ontology and reference.
Rather, these atlases are closer to those of [ADMBA](https://elifesciences.org/articles/61408), however they would also make them outdated (ADMBA's are very low resolution and hard to work with due to their undefined region boundaries).

Lastly they offer two different resolutions: one at 20µm for LSFM references and one at 31.5µm/40µm/50µm (depending on the age of the atlas) for MRI references and some stereotaxic coordinates (for post-natal atlases).
For now, in this PR, i would focus on adding just LFSM@20µm atlases.

> I am happy to give my thoughts on code quality and how to download from OneDrive

Thank you a lot!

adamltyson:
>  For now, in this PR, i would focus on adding just LFSM@20µm atlases.

That seems sensible, thanks!

carlocastoldi:
only one question remains: would you be okay if this PR also deleted the scripts for generating ADMBA and old KimP56 atlases?

They are all superseded by this new version for the developing mouse. Leaving them all would otherwise overcrowd the list of supported atlases

adamltyson:
For now, I think we should keep them. Although the new atlases may be better, there are still researchers using the old ones for ongoing projects. 

We also want to ensure that BrainGlobe acts as an archive in case the original files cannot be found for older atlases. 

Longer term, we will find a solution for making the set of available atlases easier to navigate (e.g. atlas groups).

carlocastoldi:
Sure
Thank you for the explanation!

brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_devccf_mouse.py
```diff
@@ -0,0 +1,263 @@
...
+PARALLEL = False  # disable parallel mesh extraction for easier debugging
```
alessandrofelder, line 32: As discussed in https://github.com/brainglobe/brainglobe-atlasapi/pull/441#issuecomment-2548064639, for now, I'd either
* remove the option to parallelise entirely (easier, would go for this!)
* (always parallelise in a robust way (partially addressing #429 ))

brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_devccf_mouse.py
```diff
@@ -0,0 +1,263 @@
...
+        f"In the end, {len(structures_with_mesh)} "
+        "structures with mesh are kept"
+    )
+    return meshes_dict, structures_with_mesh
+
+
+if __name__ == "__main__":
+    atlas_root = "/home/castoldi/Downloads/DevCCFv1"
+    # Generated atlas path:
+    DEFAULT_WORKDIR = Path.home() / "brainglobe_workingdir"
```
alessandrofelder, line 219: Once we figure out how to download the data, we should avoid this hard-coded path (and ideally replace it with a `pooch` temporary file)

Also, you can import the DEFAULT_WORKDIR
```python
from brainglobe_atlasapi.config import DEFAULT_WORKDIR
```
(so we can change it with one-line in the future!)

alessandrofelder (commented):
Thanks again for this, @carlocastoldi - and thank you and @adamltyson for your clarifications.

The code looks largely good on a first pass (I haven't run it, but it's promising you could use it in ABBA 🙌 ), made two small comments.

Do either of you know how official the OneDrive folder is? It looks like in the publication they point to this [Figshare repo.](https://figshare.com/articles/dataset/DevCCFv1/26377171) Using this repo instead would make it easy to use `pooch` to download and cache them neatly (and it's likely a more robust place to get data from long-term)?

carlocastoldi:
>Do either of you know how official the OneDrive folder is? 

Is the official download page from [their website](https://kimlab.io/brain-map/DevCCF/), apparently.

One that could get updated and be changed when new versions come out, I imagine.

alessandrofelder:
Hmmm, I see.

I'd still prefer going with the figshare data if possible, because
* it has a license
* it is versioned by a DOI
* (it's easier to download programmatically... haven't found a way to easily generate a sensible download link from OneDrive)

I'm not sure how standardised new versions published on OneDrive would be?
Would they make a new folder somewhere? Would they overwrite some of the existing files?
I may be missing something. 

What do you think @carlocastoldi ?

carlocastoldi:
Sure, i see all the upsides of using figshare. I guess I was worried that we wouldn't notice any update to the published atlas unless we actively search for them, but that's a minor issue!

I'll work on this PR by using `pooch` on Figshare. Thank you all!

carlocastoldi:
With this last commit, i think this PR is ready for review!

adamltyson:
Thanks @carlocastoldi. The BrainGlobe team is checked out for the holidays now, so we'll take a look in the new year.

carlocastoldi:
I figured!
I just wanted to have my part done before holidays. No hurry!

Happy holidays! ^^

brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_devccf_mouse.py
```diff
@@ -0,0 +1,242 @@
...
+    "MRI-T2",  # MRI T2-weighted
+)
+
+
+def pooch_init(download_dir_path: Path):
+    utils.check_internet_connection()
+
+    hash = None
+    registry = {a + ".zip": hash for a in TIMEPOINTS}
+    registry["DevCCFv1_OntologyStructure.xlsx"] = hash
```
alessandrofelder, line 48: It would be great if these hashes were the correct ones instead of `None`. Can be addressed separately as part of #442 if you prefer (but would be amazing if you wanted to take time for this!)

brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_devccf_mouse.py
```diff
@@ -0,0 +1,242 @@
...
+    return pooch.create(
+        path=download_dir_path,  # /ATLAS_NAME,
```
alessandrofelder, line 50:
```suggestion
        path=download_dir_path,
```

brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_devccf_mouse.py
```diff
@@ -0,0 +1,242 @@
...
+TIMEPOINTS = ("E11.5", "E13.5", "E15.5", "E18.5", "P04", "P14", "P56")
+RESOLUTIONS = (20, 50)
+MODALITIES = (
+    "LSFM",  # Light Sheet Fluorescence Microscopy
+    "MRI-adc",  # MRI Apparent Diffusion Coefficient
+    "MRI-dwi",  # MRI Difusion Weighted Imaging
+    "MRI-fa",  # MRI Fractional Anisotropy
+    "MRI-MTR",  # MRI Magnetization Transfer Ratio
+    "MRI-T2",  # MRI T2-weighted
+)
```
alessandrofelder, line 40: These are never used AFAICT - what is the intention for them? IIUC they are possible other modalities and resolutions that we don't loop over in the current version... maybe we want them in a later version?

alessandrofelder (changes requested):
Thanks a lot for this @carlocastoldi - I've tested the script locally on E11 and it's looking great 🎉 :
image

I can also confirm that the generated atlas (at least at E11) passes all our validation functions successfully.

I have a major comment that should be straightforward to handle (🤞 - happy to help)

The global variables seem to imply that the script handles all MODALITIES, but LSFM is hard-coded later on and we only loop over timepoints. We should document clearly on how the script is intended to be run. 

Maybe for this PR, just LSFM is enough?

If we wanted the other modalities (in this PR or as a separate issue for now) Maybe easiest for use is to 
    - run LSFM by default
    - have an argument to specify the modality on the command line: like `python kim_devccf_mouse.py --modality MRI-dwi`
    - replace hard-coded LSFM with modality (and `assert modality in MODALITIES`)

This would allow running atlas generation in parallel for each modality on our HPC (as separate SLURM jobs)

I have also suggested removing any obsolete (AFAICT) commented-out code.

Let me know your thoughts :)

brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_devccf_mouse.py
adamltyson: Is there any reason why we can't just add them? We have all the modalities for the P56 atlas already.

alessandrofelder: Happy to add them, no reason why not - my question was more about what the intention of this PR is, because the current code puts the modalities into a list variable that is never used?

carlocastoldi: The existence of this list was intended as ~metadata of Kim's DevCCFv1 reference images, rather than for actual generation of the atlases.

The intent of this PR was to add LSFM only because:
* its the most accurate and at highest resolution (20um vs 50 um of MRI images)
* generating 7×6=42 atlases seemed too overwhelming for an user navigating the atlases through `show_atlases()` (et simil.) output. The atlases are 7, the only thing changing is the reference image. This, in my experience, creates a sort of "bias" and confusions.

That said I think i could do adapt the script to generate those atlases as well, if you confirm me you think it's appropriate! However, i would discourage from officially adding them to BrainGlobe, until #141 is not complete.

brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_devccf_mouse.py
```diff
@@ -0,0 +1,242 @@
...
+
+if __name__ == "__main__":
+    bg_root_dir = DEFAULT_WORKDIR / ATLAS_NAME
+    download_dir_path = bg_root_dir / "downloads"
+    download_dir_path.mkdir(exist_ok=True, parents=True)
+    pooch_ = pooch_init(download_dir_path)
+    structures = fetch_ontology(pooch_)
+    # save regions list json:
+    with open(bg_root_dir / "structures.json", "w") as f:
+        json.dump(structures, f)
```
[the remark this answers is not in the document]
carlocastoldi: mmh i think i did this because i saw it was being done in other scripts as well. Is that unnecessary?

carlocastoldi:
Thank you for the comments!

I still have some questions, tho. There are parameters I was a unsure of and for which i chose the default values, when generating the meshes. Could you please help me making and informed decision:
* `closing_n_iters`
* `decimate_fraction`
* `smooth`. The resulting meshes retain the voxels shape, as we can also see from @alessandrofelder screenshot. Should we smooth them out? How do you evaluate the quality of the meshes?

Also: is it correct to use these same parameters for every age (and sizes)?
Lastly: should I add the possibility to generate the atlases for MRI@50µm reference images, could I still use the annotation volume at 20µm instead?

Thank you in advance!

carlocastoldi:
i added the `--timepoints` and `--modality` options.

This highlighted some problems when computing the meshes for the embryonic stages. _I think_ it might be due to the weird ontology/structure, but i'm not sure.
This convinces me even more to avoid to officially publish MRI atlases.

brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_devccf_mouse.py
```diff
@@ -0,0 +1,335 @@
...
+        Path(__file__).parent.parent / "hashes" / (ATLAS_NAME + ".txt")
+    )
+    return p
+
+
+def fetch_animal(pooch_: pooch.Pooch, age: str, modality: str):
+    assert age in TIMEPOINTS, f"Unknown age timepoint: '{age}'"
+    archive = age + ".zip"
+    if modality == "LSFM":
+        resolution_um = 50
```
alessandrofelder, line 64: Should this be 20?

brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_devccf_mouse.py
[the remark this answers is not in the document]
alessandrofelder: I think so
This writes a `structures.json` file to the `bg_root_dir` folder, but the actual atlases live in subfolders of that, no? The wrap-up function will [write the required structures file into each atlas's subfolder.](https://github.com/brainglobe/brainglobe-atlasapi/blob/bbc07d6a2afbc54e252dc53b8fcd8aa7f58c2ccb/brainglobe_atlasapi/atlas_generation/wrapup.py#L206)

brainglobe_atlasapi/atlas_generation/hashes/kim_devccf_mouse.txt
```diff
@@ -0,0 +1,8 @@
+DevCCFv1_OntologyStructure.xlsx 505d518b1a3b638475782150acf7750357244039585e1cc52aa48c79af005b87
```
alessandrofelder, line 1: Lovely - thank you!

alessandrofelder:
>Could you please help me making and informed decision:

    closing_n_iters
    decimate_fraction
    smooth. The resulting meshes retain the voxels shape, as we can also see from @alessandrofelder screenshot. Should we smooth them out? How do you evaluate the quality of the meshes?

The meshes are meant for visualisation only, so I would suggest smoothing them with `closing_n_iters=2` and `smooth=True` (which is the de facto default for most BrainGlobe atlases, I think).
The decimate fraction is useful if you want to reduce the size of the mesh files a bit (to improve download and rendering speeds) especially for larger, high-res atlas.
The quality "metric" is essentially: "Do the meshes look nice? (and are the mesh files not very big)?"

brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_devccf_mouse.py
carlocastoldi: yes, oh my!
Sorry for the slip-through

carlocastoldi:
>  The decimate fraction is useful if you want to reduce the size of the mesh files a bit (to improve download and rendering speeds) especially for larger, high-res atlas.

oh i understand! In light of this, do you reckon i should change the it based on the timepoint, since embryonic atlases are much smaller than post-natal?

alessandrofelder:
> oh i understand! In light of this, do you reckon i should change the it based on the timepoint, since embryonic atlases are much smaller than post-natal?

Yea, potentially... I would experiment a bit. Also, does the resolution change at later timepoints or are all LSFM atlases 20 um?

carlocastoldi:
> Also, does the resolution change at later timepoints or are all LSFM atlases 20 um?

All LSFM atlases are at 20µm. At all embryonic and postnatal stages

alessandrofelder:
> Lastly: should I add the possibility to generate the atlases for MRI@50µm reference images, could I still use the annotation volume at 20µm instead?

What would be the advantage of doing so? Are you suggesting to resample the annotation volume at 20um to 50 um? It looks like there are 50um res annotation already available?

carlocastoldi:
>What would be the advantage of doing so? Are you suggesting to resample the annotation volume at 20um to 50 um? It looks like there are 50um res annotation already available?

I guess the advantage would be that we would retain the smaller structures that otherwise would not be visible at a lower resolution. I don't know if this is the case fore Kim's, but i am sure that 25µm Allen's does not display some small regions that instead are present int 10µm annotation volume

alessandrofelder (approved):
Thanks a lot for this @carlocastoldi!
It's looking really really good!

I will run the generation of at least the LSFM atlases tomorrow (need to focus on other things today - sorry!). I'll keep this PR open until the atlases are online, so I don't forget.

What would be great is if you could open an issue
* detailing the problems you encountered with the MRI and the ontology (I'm not sure I understand that entirely)
* and any other MRI related problems (different resolution of annotations? Too many atlases?)
so we can tackle those separately to this PR (which already has a long discussion :D )

You are welcome to join our developer meeting later today (or another time) if you're free and would like to discuss the MRI side further - see [the BrainGlobe zulip ](https://brainglobe.zulipchat.com/#narrow/channel/414089-developer-meeting/topic/developer-meeting-09-01-2025/near/492366460) for details.

carlocastoldi:
>  It's looking really really good!

Thank you a lot for the speed, accuracy and time you put into reviewing this!

> I will run the generation of at least the LSFM atlases tomorrow (need to focus on other things today - sorry!). I'll keep this PR open until the atlases are online, so I don't forget.

No worries. I am not in hurry

> You are welcome to join our developer meeting later today (or another time) if you're free and would like to discuss the MRI side further - see [the BrainGlobe zulip ](https://brainglobe.zulipchat.com/#narrow/channel/414089-developer-meeting/topic/developer-meeting-09-01-2025/near/492366460)for details.

sure! It would be a pleasure. I am new to zulip, is there an already-scheduled meeting i can register to? Or will we arrange privately?
Thank you a lot! 

> What would be great is if you could open an issue

perhaps we can first discuss them in the meeting later today and evaluate if they are proper issues. If so, I will open the relative issues here, sure!

brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_devccf_mouse.py
adamltyson: I think it's appropriate. Ideally we would wait until #141 is complete, but unfortunately, if we treat #141 as a blocker, there's a lot of things we won't get done!

carlocastoldi: as you wish. I modified the script to allow other modalities as well.

Now the last thing remaining is the naming convention of the atlases. What do you prefer?
1. `kim_devccf_mouse_<MODALITY>_<AGE>`
2. `kim_devccf_mouse_<AGE>_<MODALITY>`

The 2nd option is probably  better for navigation

carlocastoldi:
with the last commit i set the atlas naming to `kim_devccf_mouse_<AGE>_<MODALITY>`.

Last thing to discuss is whether this PR should include the change of the naming to overlap previous kim atlases in brainglobe (i.e. be it `kim_dev_mouse_<AGE>_<MODALITY>`).

alessandrofelder:
> Last thing to discuss is whether this PR should include the change of the naming to overlap previous kim atlases in brainglobe (i.e. be it kim_dev_mouse_<AGE>_<MODALITY>).

I would be in favour of this last change, to make it clearer to BrainGlobe-only users that this is a newer version of the same atlas? @adamltyson may wish to chime in

adamltyson:
Yeah, if no other reason than it's a slightly shorter name.

carlocastoldi:
right.

Does that mean that I should set `VERSION=3`, since kim_dev_mouse is currently on version 2? And what about the old script? Should I delete it?

adamltyson:
> Does that mean that I should set VERSION=3, since kim_dev_mouse is currently on version 2? 

I think so

> And what about the old script? Should I delete it?

I think leave it for now, we don't (yet) have a process for this.

carlocastoldi:
I found the issue I had when generating the meshes for embryonic mri atlases.

It was pooch having an unintuitive interface, in my opinion, extracting files from archives in a different order from the one you asked.
You might have seen it working on your OS cause it depends on how the filesystem lists files. Nonetheless it wouldn't have worked on a Linux system such as Slurm, i am afraid.

With this, i feel like this PR is finally concluded! :tada:

alessandrofelder:
Thanks again for this @carlocastoldi - I am happy to say that I was able to regenerate all atlases locally and they pass our validation tests :tada: 

Upon visual inspection though, it look like all MRI reference images are converted to tiff badly. An example (E11.5, MRI-adc):

After loading the reference image with `load_nii`, we get:
```bash
# on MRI-adc data at E11.5
(Pdb) reference.dtype
dtype('float64')
(Pdb) np.max(reference)
0.024380000308156013
(Pdb) np.min(reference)
0.0
```

`reference` gets naively converted to `uint16` during [wrapup](https://github.com/brainglobe/brainglobe-atlasapi/blob/bbc07d6a2afbc54e252dc53b8fcd8aa7f58c2ccb/brainglobe_atlasapi/atlas_generation/stacks.py#L17), so all values become 0.
So I think we want to rescale the MRI data to the range [0, 2**16-1], like [we do elsewhere](https://github.com/brainglobe/brainglobe-atlasapi/blob/bbc07d6a2afbc54e252dc53b8fcd8aa7f58c2ccb/brainglobe_atlasapi/atlas_generation/atlas_scripts/axolotl_atlas.py#L112).

Does that make sense?

PS this seems to have been done in the LSFM data already, at least I checked for E11.5 programmatically (and visually they all look fine!)

```bash
# on LSFM data at E11.5
(Pdb) import numpy as np
(Pdb) np.max(reference)
65535.0
(Pdb) np.min(reference)
0.0
(Pdb) reference.dtype
dtype('float64')
```

I can take this on if you don't have time @carlocastoldi  - just let me know :)

carlocastoldi:
thanks a lot @alessandrofelder for identifying the problem and pointing me to the correct solution

carlocastoldi:
I added a `--cached-meshes` option to the script that may speed up the generation of the atlases using the same meshes (i.e. for computing meshes only once). It requires the corresponding atlas of the cached meshes to be completed, however.

**Example**
first run:
```bash
python brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_devccf_mouse.py -t E11.5 -m MRI-adc
```
and only after it finished, you can run any of:
```bash
python brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_devccf_mouse.py -t E11.5 -c MRI-adc -m MRI-dwi
python brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_devccf_mouse.py -t E11.5 -c MRI-adc -m MRI-fa
python brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_devccf_mouse.py -t E11.5 -c MRI-adc -m MRI-MTR
python brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_devccf_mouse.py -t E11.5 -c MRI-adc -m MRI-T2
```

alessandrofelder:
Thanks @carlocastoldi - I will have time to try to regenerate the MRI atlases tomorrow.

alessandrofelder:
Atlases generated locally, I will put them on GIN today - thanks again @carlocastoldi, much appreciated!

carlocastoldi:
I am sorry to come back to this, but does this work for you?

```python
import brainglobe_atlasapi
brainglobe_atlasapi.BrainGlobeAtlas("kim_dev_mouse_p04_lsfm_20um")
brainglobe_atlasapi.BrainGlobeAtlas("kim_dev_mouse_P04_LSFM_20um") # also fails to download & extract for me
```

I then run that same command in debug mode and, before running `tar = tarfile.open(destination_path)`, i swapped the `atlas.tar.gz` file with the manually downloaded archive from your Gin repo.
BrainGlobe had finally extracted the archive and created `~/.brainglobe/kim_dev_mouse_P04_LSFM_20um_v1.3`.

However, `brainglobe_atlasapi.show_atlases()` still did not show the atlas. I had to rename all-lowercase the folder of the atlas (i.e. `~/.brainglobe/kim_dev_mouse_p04_lsfm_20um_v1.3`). Only then BrainGlobe could use the ~downloaded atlas.

adamltyson:
Thanks @carlocastoldi, due to [this bug](https://github.com/brainglobe/brainglobe-atlasapi/issues/132), atlas names, and the corresponding entries in the [list here](https://gin.g-node.org/BrainGlobe/atlases/src/master/last_versions.conf) must all be lowercase. 

@alessandrofelder - might be better to update the script, then re-run all the atlas generation, rather try to rename everything (and the many errors that could occur).

PolarBean:
hey I'm just trying to figure out how to update the README to include this update. The issue i'm having is up until now I could just list the API name of each atlas (ie:  kim_dev_mouse_P04_LSFM_20um), but with this update there are now too many combinations to list all. instead of kim_dev_mouse_P04_LSFM_20um should I say kim_dev_mouse_{age}_{modality}_{resolution}um? thanks!

PolarBean:
maybe better to discuss over at https://github.com/brainglobe/brainglobe-atlasapi/pull/444

PR #444, opened 2025-01-17
document kim dev atlases in readme

alessandrofelder:

PR #447, opened 2025-01-23
update readme to include the now available ages for the kim_dev_ccf a…

PolarBean:
just updating the readme to include the ages that are provided by the kim_dev_ccf as of https://github.com/brainglobe/brainglobe-atlasapi/pull/441


adamltyson:
I think this is a duplicate of https://github.com/brainglobe/brainglobe-atlasapi/pull/444

PR #448, opened 2025-01-24
Make atlas API more robust offline

alessandrofelder:
## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**

(Possibly due to a change in requests?) our `conf_from_url` version was not working as expected, and some cases were not tested. This PR fixes this, and adds the required tests.

Also, a small fix to a previous, related test.


IgorTatarnikov (approved):
:+1: Looks good, tested by disconnecting from the internet and opening the `brainreg` napari widget. Works well!

Should we consider decreasing the number of max retries, decrease the initial sleep time, or maybe keep it flat and not increasing? It feels a bit strange launching the `brainreg` widget without internet as it hangs for 20s while the retries happen.

Issue #451, opened 2025-01-28
[BUG] Nadkarni mouse lemur regions need symmetrising


alessandrofelder:
**Describe the bug**
Enforcing https://github.com/brainglobe/brainglobe.github.io/issues/269 slipped my mind when reviewing #411 😬 I will fix and release an updated version.

alessandrofelder:
We should also add a validation function checking this.

PR #445, opened 2025-01-20
Integrating atlas validation in wrapup.py

viktorpm:
## Description

This PR integrates our atlas validation functions in the wrapup section of atlas packaging. 

**What is this PR**

- [ ] Bug fix
- [x] Addition of a new feature
- [ ] Other

**Why is this PR needed?**

Currently, we run the validation functions on finalised atlases. We want to integrate this step with atlas generation to see if an atlas is valid or not immediately.

## References

?

## How has this PR been tested?

Locally on
- whs_sd_rat.py
- princeton_mouse.py
- admba_3d_dev_mouse.py
- axolotl_atlas.py
- cavefish_atlas.py
- humanatlas.py

## Is this a breaking change?

No

## Does this PR require an update to the documentation?

Possibly

## Checklist:

- [x] The code has been tested locally
- [ ] Tests have been added to cover all new functionality (unit & integration)
- [ ] The documentation has been updated to reflect any changes
- [x] The code has been formatted with [pre-commit](https://pre-commit.com/)

PR #438, opened 2024-12-06
Hemisphere improvements to axolotl atlas

alessandrofelder:
## Description

**What is this PR**

- [x] Bug fix
- [ ] Addition of a new feature
- [ ] Other

**Why is this PR needed?**
The UNAM axolotl atlas is asymmetric and has duplicate region acronyms (because the same region exists in each hemisphere)

**What does this PR do?**
- Computes and adds the (asymmetric) hemisphere stack for the axolotl atlas
- distinguishes duplicate region acronyms by labelling them L or R

## References

First tickbox of #359 
Closes #437 

## How has this PR been tested?

Visual inspection and validation script on local axolotl atlas.

## Is this a breaking change?

No

## Does this PR require an update to the documentation?

No

PR #453, opened 2025-01-28
Add additional atlas validation functions.

alessandrofelder:
## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**What does this PR do?**
Improves our validation functions based on experiences with
* atlas names needing to be lowercase https://github.com/brainglobe/brainglobe-atlasapi/pull/446
* wrapup rescaling float64 data too naively https://github.com/brainglobe/brainglobe-atlasapi/pull/441#issuecomment-2589666369
* some atlases having differing annotations for the same region in the opposite hemisphere (e.g. #451 and #438 )

so the above issues will be flagged automatically during packaging in the future, when we merge #445 

Also, refactors our validation tests for simplicity, and makes sure we are actually testing validation functions on a valid atlas (for the tests we expect them to pass).

### DIFF
--- a/brainglobe_atlasapi/atlas_generation/validate_atlases.py
+++ b/brainglobe_atlasapi/atlas_generation/validate_atlases.py
@@ -247,6 +247,8 @@ def validate_atlas_name(atlas: BrainGlobeAtlas):
 
 
 def get_all_validation_functions():
+    """Returns all individual validation functions as a list.
+    All functions should expect 1 argument, a BrainGlobeAtlas."""
     return [
         validate_atlas_files,
         validate_mesh_matches_image_extents,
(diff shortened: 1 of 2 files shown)


alessandrofelder:
Note that one of the new validation functions catches asymmetry in labels on #451, a real-life example.
```
Failed validations:
('validate_mesh_matches_image_extents', "('Mesh coordinate 20156.5 and annotation coordinate 22932.0', 'differ by more than 10 times pixel size 91.0')")
('validate_annotation_symmetry', 'Annotation labels are asymmetric.')
```
Mesh extent validation is also expected for that atlas (there are some spurious pixels in the original data).

viktorpm (approved):

PR #449, opened 2025-01-27
Replace NIU upload action with separate actions

IgorTatarnikov:
Before submitting a pull request (PR), please read the [contributing guide](https://github.com/brainglobe/.github/blob/main/CONTRIBUTING.md).

Please fill out as much of this template as you can, but if you have any problems or questions, just leave a comment and we will help out :)

## Description

**What is this PR**

- [ ] Bug fix
- [ ] Addition of a new feature
- [x] Other

**Why is this PR needed?**
See https://github.com/neuroinformatics-unit/actions/issues/76

**What does this PR do?**
Removes the composite NIU composite action and replaces it with separate download actions/download-artifact and pypa/gh-action-pypi-publish actions.

## References
https://github.com/neuroinformatics-unit/actions/issues/76

PR #456, opened 2025-01-30
Use pypa/gh-action-pypi-publish stable release v1 

lochhh:
## Description

**What is this PR**

- [ ] Bug fix
- [x] Other

**Why is this PR needed?**
In PR #449 a specific version tag was used. We could use the release/v1 branch instead, as it will receive updates and bug fixes automatically and there's no need to update the tag as frequently. 

## References
#449 

## Checklist:

- [ ] The code has been tested locally

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -97,7 +97,7 @@ jobs:
       with:
         name: artifact
         path: dist
-    - uses: pypa/gh-action-pypi-publish@v1.12.3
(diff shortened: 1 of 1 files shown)


IgorTatarnikov (approved):
This sounds good to me, @lochhh do you think it's worth changing across all repos?

lochhh:
> This sounds good to me, @lochhh do you think it's worth changing across all repos?

Yup just wanted to start from the same first repo as the original PR.

Issue #109, opened 2021-12-06
545.obj missing in allen_mouse_25um ?


NicoKiaru:
Hi everybody, 

I don't really know where to check this, but I found a bug while trying to serialize the allen_mouse_25um_v1.2 atlas : 

`
C:\Users\nico\.brainglobe\allen_mouse_25um_v1.2\meshes\545.obj not found.
`

It seems this file is missing. Maybe that's only me ( a glitch during during the download), but maybe not ? Can you check whether this file is missing for you as well ? And if yes, can this be fixed ?

Thanks!

NicoKiaru:
Hum, it seems this obj file is not even in https://download.alleninstitute.org/informatics-archive/current-release/mouse_ccf/annotation/ccf_2017/structure_meshes/

NicoKiaru:
Maybe the issue is just that I don't know how to catch exceptions:

```
ReadError                                 Traceback (most recent call last)
~\.conda\envs\abba-bg\lib\site-packages\bg_atlasapi\structure_class.py in __getitem__(self, item)
     22             try:
---> 23                 self.data[item] = mio.read(self.data["mesh_filename"])
     24             except (TypeError, mio.ReadError):

~\.conda\envs\abba-bg\lib\site-packages\meshio\_helpers.py in read(filename, file_format)
     58         if not path.exists():
---> 59             raise ReadError(f"File {filename} not found.")
     60 

ReadError: File C:\Users\nico\.brainglobe\allen_mouse_25um_v1.2\meshes\545.obj not found.
```

vigji:
I think that the problem is what you've found, we use the meshes from the allen institute and for some reason (maybe @FedeClaudi knows more) some regions don't have one ' maybe the're too small?
What are you trying to do? Maybe you can
```
try:
    whatever(mesh)
except ReadError:
    pass
```

FedeClaudi:
I don't know why some meshes are missing! Maybe that they kept a region number from previous versions of the atlas but don't have a mesh for it any more.

Yeah the best thing you can do is to catch that error, sorry.

NicoKiaru:
Here's my issue, I need to convert a dictionary from {string: object} to {string:string}.

For that I write the following:
```
    def data(self):
        dict_ori = self.atlas.structures[self.id]
        string_dict = {key: JString(str(value)) for key, value in dict_ori.items()} #keys_values}
        return string_dict
```

[...]

PR #458, opened 2025-02-03
second version of perens lsfm atlas

PolarBean:
Alongside the perens MRI atlas they also released a new LSFM atlas which is a different space as compared to the first LSFM atlas. as it's part of the same release as the MRI atlas which was just merged this pull request is virtually identical to that one, only using different atlas files.  
https://github.com/brainglobe/brainglobe-atlasapi/pull/370

The validation runs it just complains about 3 regions being in the atlas but not having meshes. I get confused by this as in my mind there are two possiblities. 
1. The region exists in the metadata but not the segmentation file
2. The region exists in the metadata and segmentation file but is too small to make a mesh out of
If 1 is true we should remove the IDs from the metadata but if 2 is true we should not.

To smooth this process up maybe we should make distinct warnings for each case and perhaps in the case of 1 the validation script can have the option to automatically filter out regions in the metadata but not in the segmentation.

Thanks :)

### DIFF
--- a/brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_lsfm_mouse_v2.py
+++ b/brainglobe_atlasapi/atlas_generation/atlas_scripts/perens_lsfm_mouse_v2.py
@@ -1,0 +1,344 @@
+import multiprocessing as mp
+import os
+import time
+import urllib.request
+from pathlib import Path
+
+import numpy as np
+import py7zr
+from brainglobe_utils.IO.image import load_any
+from rich.progress import Progress, track
+
+from brainglobe_atlasapi import BrainGlobeAtlas, utils
+from brainglobe_atlasapi.atlas_generation.mesh_utils import (
+    Region,
+    create_region_mesh,
+)
+from brainglobe_atlasapi.atlas_generation.wrapup import wrapup_atlas_from_data
+from brainglobe_atlasapi.structure_tree_util import get_structures_tree
+
+# The Perens atlas re-uses information from the Allen atlas, so it's useful to
+# have an instance of the Allen atlas around
+allen_atlas = BrainGlobeAtlas("allen_mouse_25um")
+PARALLEL = True  # disable parallel mesh extraction for easier debugging
+
+HEADERS = {
+    "User-Agent": (
+        "Mozilla/5.0 "
+        "(Windows NT 10.0; "
+        "Win64; x64; "
+        "rv:129.0) "
+        "Gecko/20100101 "
+        "Firefox/129.0"
+    ),
+    "Accept": (
+        "text/html,"
+        "application/xhtml+xml,"
+        "application/xml;q=0.9,"
+        "image/avif,"
+        "image/webp,"
+        "image/png,"
+        "image/svg+xml,"
+        "*/*;q=0.8"
+    ),
+    "Accept-Language": "en-GB,en;q=0.5",
+    "Accept-Encoding": "gzip, deflate, br, zstd",
+    "DNT": "1",
+    "Sec-GPC": "1",
+    "Host": "www.neuropedia.dk",
+    "Connection": "keep-alive",
+    "Upgrade-Insecure-Requests": "1",
+    "Sec-Fetch-Dest": "document",
+    "Sec-Fetch-Mode": "navigate",
+    "Sec-Fetch-Site": "none",
+    "Sec-Fetch-User": "?1",
+    "TE": "trailers",
+    "Priority": "u=0, i",
+}
+
+
+# Copy-paste this script into a new file and fill in the functions to package
+# your own atlas.
+
+### Metadata ###
+
+# The minor version of the atlas in the brainglobe_atlasapi, this is internal,
+# if this is the first time this atlas has been added the value should be 0
+# (minor version is the first number after the decimal point, ie the minor
+# version of 1.2 is 2)
+__version__ = 0
+
+# The expected format is FirstAuthor_SpeciesCommonName, e.g. kleven_rat, or
+# Institution_SpeciesCommonName, e.g. allen_mouse.
+ATLAS_NAME = "perens_lsfm_mouse_v2"
+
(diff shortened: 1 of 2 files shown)


alessandrofelder (approved):
>To smooth this process up maybe we should make distinct warnings for each case and perhaps in the case of 1 the validation script can have the option to automatically filter out regions in the metadata but not in the segmentation.

This is a good idea, could you open an issue for it?
FWIW in this case I think we are in case 1, because for any of the missing regions, I get 0 for e.g. `(atlas.annotation==56/545/1051).sum()`

FWIW the Allen also has at least 545 missing #109 
I was able to generate the atlas locally and it passes the validation functions, including the new ones added in #453 

I have some reservations about having `v2` in the name of the atlas. This adds confusion. I wonder whether this script should just be `perens_lsfm_mouse.py` and continue on as a new version of the existing `perens_lsfm_mouse` atlas, but with different citation in the metadata? Thoughts appreciated @adamltyson @PolarBean

PolarBean:
yeah it is up to you guys regarding versioning but it's a tricky one. This is a distinct space from the first atlas. As we have previously discussed an atlas is composed of may things, including 
* a space
* a template
* segmentation
* ontology
Normally new atlas versions would be a new segmentation / ontology which is not really a breaking change since data anchored to the previous space would still work with the new segmentation. Here we are updating the space which is a breaking change. 

Just something to keep in mind, I'm happy for this just to become perens_lsfm_mouse

alessandrofelder:
Yes, agreed, and not sure. It's hard enough to communicate to people that there is a BrainGlobe version of an atlas, and a version of the original atlas. There is value in being explicit, but also value in simple atlas names: one would presumably check the citation of the BrainGlobe atlas and go "yes, that's the atlas I want to register my data to", cite that publication and not care whether there's a previous version of the original atlas?

adamltyson:
I think it comes down to what the "different space" is. If it's a slight change, following an improved template generation process, then I would say it's a new version of the same atlas. If it's completely different (e.g. to align with an MRI tempate) then it's a new atlas, and it should have a new name.

P.S. This is based on the BrainGlobe definition of an atlas, which so far has mostly been documented in my head. Every decision we make at the moment is a stop-gap until the atlas API V2 deals with this.

PolarBean:
I think for now we just call it `perens_lsfm_mouse`. Any other solution will be a bandaid until v2. I'll update it as soon as you guys indicate whether you agree

alessandrofelder:
IIRC we are reserving version numbers starting with 2 for atlas API v2

adamltyson:
How different is this atlas? If it varies considerably, I don't think it should be the same as the original one. 

 > I guess we should also increment the major version so this is 2.0
 
 Atlases aren't semantically versioned. The "major" version refers to the version of the API. If we update the atlas, only the "minor" version gets updated.

adamltyson:
> the second one is it's own LSFM space which is an improvement since the significant morphological distortions introduced during clearing can hamper registration.

Pretty sure that's what the first one is. It's definately not in Allen space anyway.

adamltyson:
The original atlas is definately in an LSFM space that's different to the Allen. From the paper:

>Brain regional annotations were transferred to the LSFM
template from the AIBS CCFv3 (Fig. 1b) (Allen Institute for
Brain Science 2011, 2015, 2017; Kuan et al. 2015; Wang et al.
2020). First, the mouse brain template of AIBS was registered
onto the LSFM template using multi-resolution affine and B-
spline registration. Subsequently, the registered AIBS CCFv3
template and its segmentations were divided into six parental
brain regions – cerebral cortex, cerebral nuclei, hindbrain,
cerebellum, septal regions and interbrain together with
midbrain. The parental regions were then separately reg-
istered to the corresponding areas of the LSFM tem-
plate. Manual corrections were performed for regions
near to ventricular system, such as AP and SFO.

PolarBean:
Yeah I agree. At least the first and second templates are quite different.

adamltyson:
Good point, I vote for `perens_multimodal_lsfm_25um` or `gubra_multimodal_lsfm_25um`. Probably the first to align with existing atlases.

PolarBean:
I agree I prefer `perens_multimodal_lsfm_25um`

alessandrofelder:
Also happy with ` perens_multimodal_lsfm_25um` - could you update the script accordingly @PolarBean - then I can have a look and run it locally :)

PolarBean:
done :)

alessandrofelder:
Thanks @PolarBean - I should get to this tomorrow :)

adamltyson:
>  Do we expect the perens_stereotaxic_mri_mouse_25um atlas to be in the same space as this atlas or not?

I didn't think they were, but I think @PolarBean has read the paper more recently than me!

PolarBean:
No they're not supposed to be in the same space. They're different spaces with deformation matrices defined to easily transform between them via Allen. I also think the scale difference is to be expected, I have never done lightsheet but doesn't it expand the tissue in some cases? regardless i checked the paper and it includes the following statement 

> The multimodal atlas framework includes three mouse brain templates: an MRI template, a serial two-photon microscopy (STPT) template (Wang et al., [2020a](https://link.springer.com/article/10.1007/s12021-023-09623-9#ref-CR81), [b](https://link.springer.com/article/10.1007/s12021-023-09623-9#ref-CR82)) from Mouse Brain Common Coordinate Framework version 3 (CCFv3) by Allen Institute of Brain Science (AIBS), and an LSFM template (Perens et al., [2021](https://link.springer.com/article/10.1007/s12021-023-09623-9#ref-CR54)). All maps have a isotropic voxel size of (25 μm)3.

So the voxel size we set is definitely correct! :)

adamltyson:
>  I have never done lightsheet but doesn't it chenage the size of the brain (I thought it expanded the tissue but really don't know

Varies between clearing methods. Some make it bigger, some make it smaller, some it's not so clear!

alessandrofelder:
TIL.... thanks, both!
I'll upload to GIN now.

PR #462, opened 2025-02-04
update all script names to match api names

PolarBean:
pretty small update. Just as we add more atlases its becoming confusing when the script names don't match the api names. This just updates the script names so it is consistent. If you agree we should also add this to the atlas packaging documentation.


brainglobe_atlasapi/atlas_generation/atlas_scripts/oldenburg_blackcap.py
adamltyson, line 1: This should be `eurasian_blackcap`. The atlas has been renamed (see the script itself), but maybe hasn't been uploaded to GIN yet.

adamltyson:
Seems like a good idea. Any thoughts @alessandrofelder?

brainglobe_atlasapi/atlas_generation/atlas_scripts/oldenburg_blackcap.py
PolarBean: woops my branch was a little behind

alessandrofelder:
Great idea :+1:

alessandrofelder (approved):

PR #492, opened 2025-02-05
Demba dev mouse

PolarBean:
This is to integrate the DeMBA atlases into BrainGlobe. The main thing I would like input on at this stage is the README file. Since there are 53 days of atlases each with multiple modalities it makes the atlas api table huge. What do you think we should do? I could make a html table with filters that is hosted on the brainglobe info page instead of having the table in the README. What are your thoughts?


PolarBean:
I could also make some kind of api name wizard 
image

adamltyson:
That is a lot of atlases! 

I'd rather not host an atlas-specific api name wizard if we can avoid it. Perhaps something simple like "demba_p{AGE}_{MODALITY_{RESOLUTION}um" and then a list of available ages, modalities and resolutions? 

This atlas could perhaps have a dedicated page on the website that could be linked to to outline this in more detail, and provide examples?

adamltyson:
> Annoyingly not all resolutions are available for all modalities so the demba_p{AGE}_{MODALITY}_{RESOLUTION}um may not work. 

But the available resolutions could be listed fairly succinctly right? I'm happy for the DeMBA to take up a few lines in the readme, just not 100's!

alessandrofelder:
> not all resolutions are available for all modalities so the demba_p{AGE}_{MODALITY}_{RESOLUTION}um may not work. 

Not a strong opinion, but I quite like the way the 38 `kim_dev` 1.0.0 ones are listed. (even though they take up a lot of space). Not all modalities exist for each developmental stage, so it's nice to have them listed explicitly in docs that are not meant to look as nice as our fancy webpage, and to avoid the (admittedly minimal) mental maths of "we have these 5 modalities at each developmental stage, and additionally this extra modality at E18.5 and P14 and P28" :shrug: again, not a strong opinion, happy either way

> This atlas could perhaps have a dedicated page on the website that could be linked to to outline this in more detail, and provide examples?

FWIW I like this idea

PolarBean:
I have run the validator and it works. This has the same regions as the Allen for P56 and so it's appropriate to copy the Allen ontology. Younger brains should also have the same regions (it is possible that the warping shrunk some into non existence though I haven't check for this). 

I will update the documentation and then finalise this draft pull request

image

PolarBean:
See the adjusted validation code attached here as TXT
[validation.txt](https://github.com/user-attachments/files/18708236/validation.txt)

PolarBean:
Ok I think we can start reviewing this and I'll update the documentation once we've decided how to format the table :)

alessandrofelder:
Thanks @PolarBean - I am leave most of next week, so review will have to wait until week of 17th, unfortunately.

PolarBean:
That's all good! Enjoy your leave 🏖️

adamltyson:
@carlocastoldi reminded me that atlases can have multiple reference images (https://github.com/brainglobe/brainglobe.github.io/pull/281#issuecomment-2648407455). 

For this atlas, would it be possible to resample the reference images and only create one packaged atlas per timepoint? One reference image could be picked as the "main" one, and the others added as additional references.

carlocastoldi:
I am sorry if i sort of chime here as well, but I think the ~problems i know of Kim DevCCFv1 atlases are similar to those of DeMBA.

Namely:
* the fact that at some agepoint, the total size of the reference images is small (a few hundreds MB, compressed) while at the adult stage it might be closer to be a 2+ GB, compressed
* we shouldn't just choose a "main" reference modality for each atlas. I feel like we should rank the possible modalities (e.g. 1. Average > 2. Nissl > 3. LSFM > 4. MRI) and corresponding acronyms. When creating a BG atlas, then, we would pick the first available ranked modality as the "main", and the remaining as the "additional references". Ideally, this would allow brainglobe atlases, to be ~comparable,  when possible, and use the same nomenclature

PolarBean:
I think it's a good idea to resample. But I wouldn't want to upsample only downsample. The reason is that the allen stpt is 10um resolution but the others are not.Iif I were to resample them up this would be misleading to the user as they would have no way of knowing which volumes were 'truely' a particular resolution. 

In this case we would have three resolutions

```
25um
   * allen_stpt
   * mri
   * lsfm
   * stpt

20um
   * allen_stpt
   * stpt

10um
   * allen_stpt
```

PolarBean:
In regards to the atlases being too big with too many reference volumes. I think the real solution is to only download the reference templates when they are used, instead of downloading all for a particular age.

PolarBean:
but maybe that's for V2

adamltyson:
> but maybe that's for V2

Yeah I think so, as it will require both changes to the packaging and the API

alessandrofelder:
Another limitation of the additional references is that they currently cannot be used as registration targets (IIUC, by the BrainGlobe tools). So for v1 I think we have to keep reference images that we want users to register to in separate atlases (and maybe prioritise moving to v2 rather than adding more v1 atlases altogether, including moving `brainglobe-registration` and `brainreg` to allow targetting any reference in the same atlas?)

alessandrofelder:
@PolarBean should I have a look at this yet, or would you like to implement the resampling first?

alessandrofelder:
Thanks, yea, let me know. 
If you have the latest changes from `main` in this branch, you will have the latest validation functions all integrated into the `wrapup` function directly :crossed_fingers:

brainglobe_atlasapi/atlas_generation/atlas_scripts/demba_dev_mouse.py
```diff
@@ -0,0 +1,345 @@
...
+
+    download_name = atlas_name
+    destination_path = download_dir_path / download_name
+    for url in atlas_file_url:
+        """
+        Slight issue that the hash seems different each time. I think the
+        files are being zipped on the server each time we request and it's
+        changing the hash somehow (Maybe date and time is encoded in the
+        file when zipped)
+        """
```
alessandrofelder, line 66: This is a useful comment and could be moved to the function definition for easier reading.

brainglobe_atlasapi/atlas_generation/atlas_scripts/demba_dev_mouse.py
```diff
@@ -0,0 +1,345 @@
...
+    )
+    annotation = load_any(annotation_path)
+    reference = load_any(reference_path)
+    zoom_factors = tuple(volume_resolution / resolution for _ in range(3))
+    reference = zoom(reference, zoom_factors, order=1)
+    if annotation.shape != reference.shape:
+        """
+        we unfortunately did not provide 25um segmentations so
+        we will just downsample the 20um ones
+        """
```
alessandrofelder, line 145: This could also go into its function docstring

brainglobe_atlasapi/atlas_generation/atlas_scripts/demba_dev_mouse.py
```diff
@@ -0,0 +1,345 @@
...
+    return reference, annotation
+
+
+def retrieve_additional_references(
+    download_dir_path, age, resolution, modalities
+):
+    """This function only needs editing if the atlas has additional reference
+    images. It should return a dictionary that maps the name of each
+    additional reference image to an image stack containing its data.
+    """
```
alessandrofelder, line 160: maybe a quick explanation of the relationship between the main and additional references here, instead of the boilerplate comment?

brainglobe_atlasapi/atlas_generation/atlas_scripts/demba_dev_mouse.py
```diff
@@ -0,0 +1,345 @@
...
+    If your atlas is asymmetrical, you may want to use a hemisphere map.
+    This is an array in the same shape as your template,
+    with 0's marking the left hemisphere, and 1's marking the right.
+
+    If your atlas is symmetrical, ignore this function.
+
+    Returns:
+        numpy.array or None: A numpy array representing the hemisphere map,
+        or None if the atlas is symmetrical.
+    """
```
alessandrofelder, line 188:
```suggestion
    Atlas is symmetrical, so hemisphere map is None.

    Returns:
        None: Atlas is symmetrical
    """
```

brainglobe_atlasapi/atlas_generation/atlas_scripts/demba_dev_mouse.py
```diff
@@ -0,0 +1,345 @@
...
+
+def retrieve_structure_information():
+    """
+    Retrieve the structures tree and meshes for the Allen mouse brain atlas.
+
+    Returns:
+        pandas.DataFrame: A DataFrame containing the atlas information.
+    """
+    # Since this atlas inherits from the allen can we not simply get the data
+    # from the bgapi?
```
alessandrofelder, line 200:
```suggestion
```
Yes! and we do. Nice!

brainglobe_atlasapi/atlas_generation/atlas_scripts/demba_dev_mouse.py
```diff
@@ -0,0 +1,345 @@
...
+        if key in labels:
+            is_label = True
+        else:
+            is_label = False
+
+        node.data = Region(is_label)
+
+    # Mesh creation
+    closing_n_iters = 2  # not used for this atlas
+    decimate_fraction = 0.2  # not used for this atlas
```
alessandrofelder, line 240:
```suggestion
    closing_n_iters = 2
    decimate_fraction = 0.2
```
Comments not true, I think?

brainglobe_atlasapi/atlas_generation/atlas_scripts/demba_dev_mouse.py
```diff
@@ -0,0 +1,345 @@
...
+
+    print(
+        f"In the end, {len(structures_with_mesh)} "
+        "structures with mesh are kept"
+    )
+    return meshes_dict
+
+
+### If the code above this line has been filled correctly, nothing needs to be
+### edited below (unless variables need to be passed between the functions).
```
alessandrofelder, line 286:
```suggestion
```
Comment confusing. We have custom code below iterating over the ages of the atlas.

brainglobe_atlasapi/atlas_generation/atlas_scripts/demba_dev_mouse.py
```diff
@@ -0,0 +1,345 @@
...
+                    bg_root_dir, age, resolution, modalities[0]
+                )
+            )
+            if len(modalities) > 1:
+                additional_references = retrieve_additional_references(
+                    bg_root_dir, age, resolution, modalities[1:]
+                )
+            hemispheres_stack = retrieve_hemisphere_map()
+            structures = retrieve_structure_information()
+            if meshes_dict is None:
```
alessandrofelder, line 312:
```suggestion
           # generate pixel-scale mesh files only once, for 25um, and 
           # re-use them and the meshes_dict
            if meshes_dict is None:
```
Is what I think is happening?

alessandrofelder (commented):
I have had a look at the code, and it looks really nice!
I made some nitpicky comments that won't affect the functionality but IMO improve code legibility.

I am running this locally now too, and will approve when validation passes :)

brainglobe_atlasapi/atlas_generation/atlas_scripts/demba_dev_mouse.py
PolarBean: this is from the template script. Ill remove it

brainglobe_atlasapi/atlas_generation/atlas_scripts/demba_dev_mouse.py
PolarBean: good suggestion

brainglobe_atlasapi/atlas_generation/atlas_scripts/demba_dev_mouse.py
PolarBean: Ok potential problem. This atlas actually uses the Allen 2022 segmentations. The Allen in brainglobe uses 2017. Will this cause an issue?

PolarBean: I think yes

brainglobe_atlasapi/atlas_generation/atlas_scripts/demba_dev_mouse.py
PolarBean: correct!

brainglobe_atlasapi/atlas_generation/atlas_scripts/demba_dev_mouse.py
PolarBean: This code has been copy and pasted from one script to another like ten times now. It is both cursed and haunted by comments from atlases past.

brainglobe_atlasapi/atlas_generation/atlas_scripts/demba_dev_mouse.py
adamltyson: In general ,I think it's probably best not to rely on one atlas to build another. We could get into a situation where one atlas changes, and we're unable to build another.

alessandrofelder: OK, in that case we should fix this before merging.

PolarBean: I fixed it :)

PolarBean:
FYI the validation worked. See the updated validation script attached as a txt. [validation.txt](https://github.com/user-attachments/files/18870038/validation.txt)
image

PolarBean:
We have another version of this atlas which uses the kim developmental segmentations. I'm going to open a second pull request once this one is merged for it. Should I rename this to be demba_dev_allen_seg_mouse and have that one as demba_dev_kim_seg_mouse ?

adamltyson:
> Should I rename this to be demba_dev_allen_seg_mouse and have that one as demba_dev_kim_seg_mouse ?

Sounds good to me.

alessandrofelder:
Happy with suggested names too

@PolarBean would you be able to share the tar.gz files directly when you're done?
Maybe even including the renaming?
Seems more efficient for this scale of atlas packaging than me regenerating them locally (twice) :grin:

PolarBean:
Yeah I can do that :)

alessandrofelder:
(I can merge this and wait for your follow-up renaming PR if that's helpful?)

PolarBean:
How about we do the renaming before merging. I like the idea of rarely renaming atlases once they are merged.

alessandrofelder (approved):

PolarBean:
https://data-proxy.ebrains.eu/deepslice?prefix=bg_sharing%2F

Here are some of the ages generated. Let me know if they're all good and ill generate the rest :)

PolarBean:
Should be no worries to run one or two!

IgorTatarnikov:
I've downloaded the atlases and run validation on the local copies. I'll attach the results here [demba_validation_log.txt](https://github.com/user-attachments/files/22233009/demba_validation_log.txt).

In summary:

Several atlases have missing meshes:
- Based on the discussion, that's expected due to the small size of some regions
- The missing areas are some combination of `[164, 375, 382, 473, 545, 1051]`. 

`validate_mesh_matches_image_extents` failures:
- p26_20um, p33_20um, p33_25um, p45_20um, p45_25um, p48_20um, p56_20um, p56_25_um fail `validate_mesh_matches_image_extents`
- Return some form of `('Mesh coordinate 7530.0 and annotation coordinate 7980.0', 'differ by more than 10 times pixel size 20.0')`

`validate_annotation_symmetry` failures:
- Some atlases fail the annotation label symmetry, from checking several of these I think the L/R hemisphere split is close to the middle plane which means they can be labelled as `symmetric` in BrainGlobe without consequences.

If these are expected and not blocking, I can upload the atlases to GIN and merge this PR.

IgorTatarnikov:
Doing some digging, p56_20um has a couple of stray pixels in the annotations. At `(659, 399, 260)` and `(659, 399, 309)` the labels are 471, when they should be 0. 

image

adamltyson:
I think that's fine then, unless @PolarBean wants to update the originals? If not, we can package these and update them down the line if needed.

IgorTatarnikov:
Sounds good! For future debugging purposes, the pixels always seems to be at the posterior end, sitting the maximum y coordinate, and always have the label 471 which doesn't correspond to any structure (I think).

I'll upload the atlases to GIN now.

The packaging script needs to be moved into the root `atlas_scripts` directory prior to merging as per https://github.com/brainglobe/brainglobe-atlasapi/pull/625.

IgorTatarnikov:
All atlases have been uploaded and are available for BrainGlobe tools to use!

PolarBean:
Great, ill fix those docstrings asap

PR #496, opened 2025-02-07
Table cleanup

PolarBean:
I want to keep a consistent look for the table and I feel the list of api names is perhaps becoming a bit unwieldy. Especially with the introduction of DeMBA. I also think that we should somewhat standardise it so having one row for DeMBA and multiple rows for DevCCF might be a bit untidy. Here I propose having a button to redirect towards the docs. What are your thoughts @alessandrofelder @adamltyson


adamltyson:
I like the idea a lot, but at least on the machine I'm using the `More info | Click here` buttons are very small (smaller than the normal text). Is there a way to make these larger/clearer?

Alsom would you mind removing all the demba packaging code from this PR?

PolarBean:
I made them bigger, slightly hacky i used an invisible unicode character instead of a space to create a wider column. The reason i did it this way is that github doesn't support custom CSS. Sorry about DeMBA packaging code it slipped my notice. I guess I created this branch while in the DeMBA branch!

adamltyson:
Looks good to me, certainly an improvement anyway.

Any thoughts @alessandrofelder?

alessandrofelder:
Very nice!

adamltyson:
@PolarBean is this ready to merge (it's still in draft)?

PolarBean:
yeah it's ready

PR #498, opened 2025-02-11
Adds fixture to download atlases for test_validation.py

IgorTatarnikov:

## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**
Tests were failing as `test_validation.py` tried to modify a directory that didn't exist. 

https://github.com/brainglobe/brainglobe-atlasapi/actions/runs/13244823116/job/36975288009?pr=497

This failure seems to be related to the CI cache miss.

**What does this PR do?**
Ensures atlases are downloaded prior to `test_validation.py` running.

All tests pass locally and on CI.

## Is this a breaking change?
No

No

## Checklist:

### DIFF
--- a/tests/atlasgen/conftest.py
+++ b/tests/atlasgen/conftest.py
@@ -1,0 +1,21 @@
+from pathlib import Path
+
+import pytest
+
+from brainglobe_atlasapi import BrainGlobeAtlas
+
+
+@pytest.fixture(autouse=True)
+def setup_preexisting_local_atlases():
+    """Automatically setup all tests to have three downloaded atlases
+    in the test user data."""
+    preexisting_atlases = [
+        ("example_mouse_100um", "v1.2"),
+        ("allen_mouse_100um", "v1.2"),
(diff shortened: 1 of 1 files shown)


niksirbi (approved):
Looks good Igor.

There's one thing I'm not sure about (because I don't know the working of atlasapi very well):
Here you are checking if 3 atlases of specific version exist, but upon download (`BrainGlobeAtlas` instantiation), you don't actually enforce the downloading of the specific versions you checked. I assume the latest version will be fetched instead. Will there be a problem if the specific versions you check are no longer the latest ones?

IgorTatarnikov:
Thanks!

The code in `test_validation.py` also looks for specific atlas versions when looking in the directory tree, see: https://github.com/brainglobe/brainglobe-atlasapi/blob/4b5fa6479a0cf209a15f853cf48be6fa33f9bb6e/tests/atlasgen/test_validation.py#L32C1-L44C35.

I think that's something we should discuss at some point, do we run tests on specific versions or the latest?

Issue #459, opened 2025-02-03
[Feature] Standardise naming scheme


PolarBean:
We should consistently follow a naming convention for API names
eg; (author or atlas name)_(species)_(age)_(modality)_(resolution)um

we sometimes mix the order (eg; nadkarni_mri_mouselemur_91um vs perens_stereotaxic_mouse_mri_25um)

I think we should prioritise not mixing the order but another discussion is whether fields can be left blank. At present it seems we often don't mention modality (eg; allen_mouse_10um instead of allen_mouse_stpt_10um)

adamltyson:
> We should consistently follow a naming convention for API names
eg; (author or atlas name)(species)(age)(modality)(resolution)um

Agree

> I think we should prioritise not mixing the order but another discussion is whether fields can be left blank. At present it seems we often don't mention modality (eg; allen_mouse_10um instead of allen_mouse_stpt_10um)

I think things can be left out, as this meant to be a name, not comprehensive metadata. There's only one allen mouse atlas, so we don't need to specify the modality. Of course, detailed metadata should be added to the docs, and eventually added to the atlas itself.

PolarBean:
we would also need to mention organ. Perhaps the assumption is that it is brain and we specify if it is anything else (eg; cord)

adamltyson:
> what about if an atlas has a two word name like perens stereotaxic? should we enforce one word names?

I think multiple is fine, we shouldn't be using these names to parse metadata. 

> we would also need to mention organ. Perhaps the assumption is that it is brain and we specify if it is anything else (eg; cord)

Yeah, an optional entry after species?

PolarBean:
Since this would be a breaking change is there anything in particular we should do before implementing it. Or is it something that we should not apply retroactively?

adamltyson:
TBH, I would say this should be a (documented) guide for future atlases. I wouldn't say it's high priority enough to start changing every existing atlas.

PolarBean:
We also need to specify how to write decimal points in the api name. At the moment there are 3 ways. 

- `kim_dev_mouse_e11-5_mri-adc_31.5um` - uses a dash for 11.5 but uses a decimal point for 31.5
- `admba_3d_e11_5_mouse_16um` uses an underscore

adamltyson:
This was always meant to be underscores for everything. More things to document and add to the validator ...

PolarBean:
[...]

PR #500, opened 2025-02-14
Allen bbp ccfv3a entire mouse brain, improve mesh utils, and add metadata filtering to wrapup

PolarBean:
Here I add the extended and improved CCFv3 annotation and Nissl atlas of the entire mouse brain by Piluso and colleagues (full disclosure I'm also an author). 

I have also integrated the filtering script we have previously talked about that automatically removes metadata regions that do not exist in the annotation, previously a point of friction in integration scripts. I moved int into the wrapup script. 

Another addition here is the addition of a annotation + region dictionary to mesh function. This code is the same in every script I submit and is a large effectively boiler plate function that is copy and pasted between them. By moving it into mesh utils we make atlas integration easier and the integration scripts cleaner.


PolarBean:
given that this relies on changes to mesh utils you need to `pip install --upgrade -e .` in the main brainglobe directory before running the integration script

PolarBean:
Found an error. Closed until I fix

PolarBean:
Ok fixed some orientation problems, but another issue was the datatype. It seems all brainglobe atlases must be 16 bit unsigned integers. Is this in the integration docs? definitely something to add, not sure where though. 

Ive also started using dataclasses here, I was thinking about upstreaming them into brainglobe and importing the atlas_metadata class in future atlas requests. 

Theres this dictionary here that could be replaced by a dataclass if youd like and imported as needed https://github.com/brainglobe/brainglobe-atlasapi/blob/main/brainglobe_atlasapi/descriptors.py

adamltyson:
> Ok fixed some orientation problems, but another issue was the datatype. It seems all brainglobe atlases must be 16 bit unsigned integers. Is this in the integration docs? definitely something to add, not sure where though.

Don't think it's documented anywhere. It should probably go somewhere [here](https://brainglobe.info/documentation/brainglobe-atlasapi/adding-a-new-atlas.html#atlas-scripts).

brainglobe_atlasapi/atlas_generation/atlas_scripts/allen_bbp_ccfv3a_entire_mouse_brain.py
```diff
@@ -0,0 +1,267 @@
...
+
+# Copy-paste this script into a new file and fill in the functions to package
+# your own atlas.
+
+### Metadata ###
+
+# The minor version of the atlas in the brainglobe_atlasapi, this is internal,
+# if this is the first time this atlas has been added the value should be 0
+# (minor version is the first number after the decimal point, ie the minor
+# version of 1.2 is 2)
```
alessandrofelder, line 27:
```suggestion
```

brainglobe_atlasapi/atlas_generation/atlas_scripts/allen_bbp_ccfv3a_entire_mouse_brain.py
```diff
@@ -0,0 +1,267 @@
...
+METADATA = AtlasMetadata(
+    version=0,
+    name="allen_bbp_ccfv3a_entire_mouse_brain",
+    citation="https://doi.org/10.1101/2024.11.06.622212",
+    atlas_link="https://zenodo.org/api/records/14034334/files-archive",
+    species="Mus musculus",
+    orientation="spr",
+    root_id=997,
+    resolution=25,
+)
```
alessandrofelder, line 69: Not a strong opinion, but maybe a worry would be that using a dataclass (while nice syntactic sugar for more experienced developers) introduces a small barrier for novice programmers to contribute atlases in the future (because they need to understand what `e.g. METADATA.name` refers to)?

brainglobe_atlasapi/atlas_generation/atlas_scripts/allen_bbp_ccfv3a_entire_mouse_brain.py
```diff
@@ -0,0 +1,267 @@
...
+
+def retrieve_or_construct_meshes(download_path, annotated_volume, structures):
+    """
+    This function should return a dictionary of ids and corresponding paths to
+    mesh files. Some atlases are packaged with mesh files, in these cases we
+    should use these files. Then this function should download those meshes.
+    In other cases we need to construct the meshes ourselves. For this we have
+    helper functions to achieve this.
+    """
+    meshes_dict = construct_meshes_from_annotation(
```
alessandrofelder, line 217: yea, this is super-useful as a helper function :star_struck:

brainglobe_atlasapi/atlas_generation/mesh_utils.py
```diff
@@ -245,6 +247,86 @@ def create_region_mesh(args):
...
+
+    tree = get_structures_tree(structures_list)
+    labels = np.unique(volume).astype(np.int32)
+
+    for key, node in tree.nodes.items():
+        node.data = Region(key in labels)
+
+    closing_n_iters = 2
+    decimate_fraction = 0
+    smooth = False
```
alessandrofelder, line 287: I wonder whether these should be exposed as parameters, with default values, so we can re-use this function across more packaging scripts (as and when we need to, moving to version 2)

brainglobe_atlasapi/atlas_generation/wrapup.py
```diff
@@ -116,6 +136,11 @@ def wrapup_atlas_from_data(
 
     # If no hemisphere file is given, assume the atlas is symmetric:
     symmetric = hemispheres_stack is None
+    if isinstance(annotation_stack, str) or isinstance(annotation_stack, Path):
+        annotation_stack = tifffile.imread(annotation_stack)
```
alessandrofelder, line 140: Why is this needed? Hopefully we'd always read the annotation into an array in `retrieve_reference_and_annotation`

brainglobe_atlasapi/atlas_generation/atlas_scripts/allen_bbp_ccfv3a_entire_mouse_brain.py
```diff
@@ -0,0 +1,267 @@
...
+            bg_root_dir, resolution=resolution
+        )
+
+        hemispheres_stack = retrieve_hemisphere_map()
+        if resolution == 25:
+            structures = retrieve_structure_information(bg_root_dir)
+            meshes_dict = retrieve_or_construct_meshes(
+                bg_root_dir, annotated_volume, structures
+            )
+        reference_volume = (reference_volume * 65535).astype(np.uint16)
```
alessandrofelder, line 247: presumably you're sure the reference volume pixels are all between 0.0 and 1.0 here?

alessandrofelder (changes requested):
Thanks a lot @PolarBean! Generally, looks very nice, in particular the additional helper functions :raised_hands: 

Could you remove left-over docstrings from the template script, and replace them with descriptions of this atlases idiosyncrasies (many of which you already detail in code comments), please?

Then I can run and validate locally too :)

brainglobe_atlasapi/atlas_generation/mesh_utils.py
PolarBean: absolutely

brainglobe_atlasapi/atlas_generation/wrapup.py
PolarBean: We need annotation stack loaded as an array for filter_structures_not_present_in_annotation
https://github.com/brainglobe/brainglobe-atlasapi/blob/d9124c690d9e6f2c8caa58a5067b8006e31df3b9/brainglobe_atlasapi/atlas_generation/wrapup.py#L141-L143

brainglobe_atlasapi/atlas_generation/atlas_scripts/allen_bbp_ccfv3a_entire_mouse_brain.py
PolarBean: Yeah :)

brainglobe_atlasapi/atlas_generation/mesh_utils.py
PolarBean: I exposed these parameters. I havent tested it yet as im currently generating demba files (and will be for a while!). There is the option to expose all of the parameters of extract mesh from mask but I'm not to familiar with that function.

brainglobe_atlasapi/atlas_generation/atlas_scripts/allen_bbp_ccfv3a_entire_mouse_brain.py
PolarBean: Interesting point, I don't think its so difficult for novices but its hard to say. I think the main point of this change is to organise all the metadata in one place with detailed descriptions of what each field requires and what type of data it expects. I think doing this with constants and comments in a template file is less than ideal perhaps but I'm open to opinions.

brainglobe_atlasapi/atlas_generation/wrapup.py
alessandrofelder: Ah, yea, I see. Maybe in the future we want to disallow passing a path or string in the first place, but that's outside the scope of this PR.

brainglobe_atlasapi/atlas_generation/atlas_scripts/allen_bbp_ccfv3a_entire_mouse_brain.py
alessandrofelder: yea, it's a trade-off. Not sure - will bring up at dev meeting tomorrow.

adamltyson: I would vote against dataclasses. Atlas contributors tend to be the least experienced contributors we have (some have no Python experience). Anything we can do to simplify this process is good.

PolarBean: fair point! should I remove it from this script or is it ok

PolarBean:
it runs :)

brainglobe_atlasapi/atlas_generation/mesh_utils.py
PolarBean: it runs :)

brainglobe_atlasapi/atlas_generation/atlas_scripts/allen_bbp_ccfv3a_entire_mouse_brain.py
alessandrofelder: Let's remove it for consistency across packaging scripts?

alessandrofelder:
We should also think about testing the helper functions (here or in a different PR)

PolarBean:
I don't think I have the bandwidth at the moment aha sorry

PolarBean:
This is ready for review :)

alessandrofelder (commented):
At least for the versions I generated locally:

It looks like there's a mismatch in the mesh dimensions in the the 10um version?

image
which is also flagged by the validation:
```bash
Failed validations:
('validate_mesh_matches_image_extents', "('Mesh coordinate 4985.0 and annotation coordinate 0.0', 'differ by more than 10 times pixel size 10.0')")
```

The dimensions match for 25um, 
image

and it looks to me like the 10um version is not scaled - I don't see from the code why that happens though. Can you reproduce @PolarBean ?


There also seem to be some extra meshes?

```bash
('catch_missing_structures', 'Structures with IDs [991, 254, 909, 567, 714, 669, 453, 792, 1008, 528, 500, 95, 315, 247, 22, 322, 31] have a mesh file, but are not accessible through the atlas.')
```

These are presumably expected, as in the Allen atlas, e.g #109
```bash
('catch_missing_mesh_files', "Structures with IDs [545, 1051] are in the atlas, but don't have a corresponding mesh file.")
```
 
It also looks like the resulting BrainGlobe orientation is not the expected ASR (should open as a coronal section in napari by default).
image

PolarBean:
Yeah the meshes not being scaled is because I tried saving time using the 25um meshes for both. I don't know why but I thought brainglobe atlases did that. It might be what we do internally and I had crossed wires. I thought I had fixed the issue with removing too many regions from the metadata.... let me work on that and then rerun the validation....

PolarBean:
It was only checking if an id or its children were in the annotation, I needed to check whether an id or any of its descendants were in the annotationand only if not then discard it from the metadata. There were some regions that werent in the annotationbut had grandchildren and great grandchildren in the annotation.

PolarBean:
I've run into an annoying issue and it turns out the issue is with the Allen atlas. 
This is the mesh for region 1051 
image
However when I download the annotation (2017) from the allen and quantify the region volumes it tells me this region is 4 voxels at 10um. 
image
The only way this is possible I suppose is because the brain globe integration script doesnt calculate the meshes from the volume but rather directly downloads them from the Allen. 

So the issue is that the annotation and meshes released by the allen are out of sync and they have not properly released the annotation for region 1051 (direct tectospinal pathway). 

Let me know if you can think of another explanation. area 545 (Retrosplenial area, dorsal part, layer 4) is only 2 voxels but also has no corresponding mesh so isn't an issue. 

Besides this issue for some reason I cannot get the validation to pass at 10um. Napari throws this error and I really don't know where to start with it. 
image

It works just fine at 25um so I'm really puzzled. I have no experience with Napari so would really appreciate some advice here. Here are the 25um and 10um datasets in case you'd like to have a look but don't have time to run the whole script yourself (it takes more than 24 hours). 
https://data-proxy.ebrains.eu/api/v1/buckets/deepslice/allen_bbp_ccfv3a_entire_mouse_brain_25um_v1.0.tar.gz
https://data-proxy.ebrains.eu/api/v1/buckets/deepslice/allen_bbp_ccfv3a_entire_mouse_brain_10um_v1.0.tar.gz

PolarBean:
The files themselves look correct so it may be an issue with my napari installation

alessandrofelder:
Thanks for excellent detective work, @PolarBean 

> So the issue is that the annotation and meshes released by the allen are out of sync and they have not properly released the annotation for region 1051 (direct tectospinal pathway).
> 
> Let me know if you can think of another explanation. area 545 (Retrosplenial area, dorsal part, layer 4) is only 2 voxels but has no corresponding mesh so isn't an issue.

This is deeper than we've gone before, but IIRC we are aware of those two areas being "weird" in the Allen (right, @viktorpm ?)

I am at a conference all week, so I've put a thorough review of this on my list for Monday next week.
I can help with `napari` in the meantime, though... what exactly are you doing to cause that error? (and yes, re-installing `napari` might help)

viktorpm:
Yes, we catch 545 during validation but I don't remember any issue with 1051.

PolarBean:
> I can help with napari in the meantime, though... what exactly are you doing to cause that error? (and yes, re-installing napari might help)
Just running the validation script. It launches the 25um no problem its only an issue with 10um. I'll troubleshoot some more.

PolarBean:
Ok it seems like an out of memory issue as napari tries to create a 3D projection of the annotation instead of using the meshes? In that case theres nothing wrong with the atlas as it is... below is the validation code I'm using I would appreciate you running it when you get time. Thanks a lot. 
```python
import shutil
from pathlib import Path

import napari
from brainrender_napari.napari_atlas_representation import (
    NapariAtlasRepresentation,
)
from napari.viewer import Viewer

from brainglobe_atlasapi import BrainGlobeAtlas
from brainglobe_atlasapi.atlas_generation.validate_atlases import (
    catch_missing_mesh_files,
    catch_missing_structures,
    open_for_visual_check,
    validate_additional_references,
    validate_atlas_files,
    validate_checksum,
    validate_image_dimensions,
    validate_mesh_matches_image_extents,
)

all_validation_functions = [
    validate_atlas_files,
    validate_mesh_matches_image_extents,
    open_for_visual_check,
    validate_checksum,
    validate_image_dimensions,
    validate_additional_references,
    catch_missing_mesh_files,
    catch_missing_structures,
]


# adapt this code block for newly packaged atlases
brainglobe_dir = Path.home() / ".brainglobe/"
working_dir = Path.home() / "brainglobe_workingdir"
atlas_name = "allen_bbp_ccfv3a_entire_mouse_brain"
resolution = 10
minor_version = 0

# nothing below this needs changing

# make sure we have the latest packaged version in .brainglobe
# by replacing it with the working_dir version if needed
atlas_name_with_version = (
    f"{atlas_name}_{resolution}um_v1.{minor_version}"
)
source_dir = working_dir / atlas_name / atlas_name_with_version
destination_dir = brainglobe_dir / atlas_name_with_version
if destination_dir.exists() and destination_dir.is_dir():
    shutil.rmtree(destination_dir)
print(source_dir)
assert source_dir.exists()
if source_dir.exists():
    shutil.copytree(source_dir, destination_dir)
assert destination_dir.exists()

# run validation functions on the new atlas
atlas = BrainGlobeAtlas(f"{atlas_name}_{resolution}um")
validation_results = {atlas_name: []}

for i, validation_function in enumerate(all_validation_functions):
    try:
        validation_function(atlas)
        validation_results[atlas_name].append(
            (validation_function.__name__, None, str("Pass"))
        )
    except AssertionError as error:
        validation_results[atlas_name].append(
            (validation_function.__name__, str(error), str("Fail"))
        )

# print validation results and open napari for a visual check
# in napari, we should see three layers:
# - the annotation
# - the reference image (visibility turned off by default)
# - the root mesh

failed_validations = [
    (result[0], result[1])
    for result in validation_results[atlas_name]
    if result[2] == "Fail"
]
if failed_validations:
    print("Failed validations:")
    for failed in failed_validations:
        print(failed)
else:
    print(f"{atlas_name} is a valid atlas")

viewer = Viewer()
viewer.dims.ndisplay = 3
napari_atlas = NapariAtlasRepresentation(atlas, viewer)
napari_atlas.add_structure_to_viewer("root")
napari_atlas.add_to_viewer()

napari.run()

```

alessandrofelder:
Ah, we have built validation into the wrapup function now, so the packaging script should print any errors it notices as you run it - sorry, as usual: this needs better docs. Making time for that is tricky, but I am trying.

We have also simplified post-hoc validation:
* you can get a list of all validation functions
* they work with atlases that are not necessarily in the default folder

This allows you to create little scripts like this:
```python
from brainglobe_atlasapi.atlas_generation.validate_atlases import get_all_validation_functions
from pathlib import Path
from brainglobe_atlasapi import BrainGlobeAtlas

atlas = BrainGlobeAtlas("allen_bbp_ccfv3a_entire_mouse_brain_10um", brainglobe_dir=Path.home()/"Downloads/brainglobe-test", check_latest=False)

for function in get_all_validation_functions():
    try:
        print("-----")
        print(f"{function.__name__} running")
        function(atlas)
    except Exception as e:
        print(f"{function.__name__} caught error")
        print(e)
    print("------")
```
and separated it from visualisation
```python
from pathlib import Path
from brainglobe_atlasapi import BrainGlobeAtlas
from brainrender_napari.napari_atlas_representation import NapariAtlasRepresentation
import napari

atlas = BrainGlobeAtlas("allen_bbp_ccfv3a_entire_mouse_brain_10um", brainglobe_dir=Path.home()/"Downloads/brainglobe-test", check_latest=False)
viewer = napari.Viewer()
napari_atlas = NapariAtlasRepresentation(atlas, viewer)
napari_atlas.add_to_viewer()
napari_atlas.add_structure_to_viewer("root")

napari.run()
```

(Maybe we should make these scripts into napari tools 🤔 )

The good news is, for the 10um atlas that you shared, I get only one unexpected validation error (the 545 one is expected): it looks like the root mesh and the annotations don't match so well, in one dimension?

```
-----
validate_atlas_files running
------
-----
validate_mesh_matches_image_extents running
validate_mesh_matches_image_extents caught error
('Mesh coordinate 7545.0 and annotation coordinate 7990.0', 'differ by more than 10 times pixel size 10.0')
------
-----
open_for_visual_check running
------
-----
validate_checksum running
------
-----
validate_image_dimensions running
------
-----
validate_additional_references running
------
-----
catch_missing_mesh_files running
catch_missing_mesh_files caught error
Structures with IDs [545] are in the atlas, but don't have a corresponding mesh file.
------
-----
catch_missing_structures running
------
-----
validate_reference_image_pixels running
------
-----
validate_annotation_symmetry running
------
-----
validate_atlas_name running
------
```

The prime suspect is a spurious pixel in the annotations. Let's check by visualising:

image

generally looks good ⬆️ 
They are tricky to find, but it looks like on slice 1354 of the annotation image, below the medulla, there's some loose pixels that think they are part of CA3 (they have value 471).

image
 
As discussed previously, this is fine - as long as we know it's a consequence of us wanting to conserve the original data. Is this the case?

PolarBean:
haha wow yeah, good catch. I checked and it is in the atlas file shared by the authors. I wasn't involved in the creation of the delineations and since the blue brain has now ended I don't think they would be able to remove the voxel. In this case I would say share as is and it can be corrected in the future brainglobe cleaned atlases we have discussed.

alessandrofelder:
Thanks for confirming @PolarBean - I will run the validation on the 25um version next week, but I expect it will pass too (with the two expected errors), and if so, will upload to GIN.

Are we happy with the name `allen_bbp_ccfv3a_entire_mouse_brain`?
And would you have capacity for making PRs to update the docs accordingly by any chance?

PolarBean:
Im happy with the name though it is long. Do you have another name you'd prefer? I can update the docs :)

PolarBean:
ccfv3a_entire_mouse_brain?

alessandrofelder:
Hmmm after re-reading our discussion in https://github.com/brainglobe/brainglobe-atlasapi/issues/459 I'd be tempted to leave out the `brain` (implied) and maybe split the "entire/augmented" descriptor from the species, to follow `(author or atlas name)\_(species)\_(age)\_(modality)\_(resolution)um`... so my suggestion is

`bluebrain_ccfv3augmented_mouse_10um`, also making it (kind of) consistent with the BlueBrain barrel atlas that we already have?

alessandrofelder:
or `bluebrain_augmented_mouse_10um` or `bluebrain_augmented_allen_mouse_10um` 🤔

PolarBean:
I like bluebrain_ccfv3augmented_mouse_10um the best :)

PolarBean:
to be more consistent with the proposed standardised naming scheme of 
`(author or atlas name)\_(species)\_(age)\_(modality)\_(resolution)um`
should we say ccfv3augmented_mouse_nissl_10um 
I'll regenerate it as I think I want the population average to be the default template.

adamltyson:
I would say miss out nissl as it's the stain, or add the imaging modality.

PolarBean:
I think "Nissl volume" is the term for data like these that I see used most commonly so that's probably the most informative name (for an example see the second [paragraph here](https://download.alleninstitute.org/informatics-archive/october-2014/mouse_expression/Accessing_October_2014_expression_data.pdf)). The volume itself is reconstructed from histological sections but histology can vary in appearance so much (I can imagine a ccf_mouse_myelin).

Which would you prefer  `ccfv3augmented_mouse_nissl_10um` or `ccfv3augmented_mouse_histology_10um` or just `ccfv3augmented_mouse_10um`. I vote either the first or last but any is fine.

adamltyson:
I would go for the last, because I think our proposed standard including "modality" needs some refinement.

PolarBean:
great I'll generate it now :)

brainglobe_atlasapi/atlas_generation/atlas_scripts/validation.py
```diff
@@ -0,0 +1,98 @@
+import shutil
```
alessandrofelder, line 1: Looks like this file is a local validation script for a different atlas, and probably added to this PR in error?

alessandrofelder (commented):
Thanks @PolarBean 

I confirm the data you shared passes validation (apart from the expected two cases at 10um, as discussed above) and generally, changes are sensible. (there's a demba validation file that is unrelated to the PR)

Final check though from manual visualisation: do we expect the background pixels of the main reference image to be the brightest? It looks different than the publication - I suspect the culprit is your multiplication by 65535?

(IIUC and the intention is to rescale the pixel range to 0-65535 you would need to subtract the minimum and scale by difference of max and min first - to force the range to be 0.0-1.0 - and then multiply?)

brainglobe_atlasapi/atlas_generation/atlas_scripts/validation.py
PolarBean: yes apologies, I'll remove it

PolarBean:
I multiplied it by 65535 twice 🤦

alessandrofelder:
Thanks @PolarBean - I should have time to double-check this and hopefully upload tomorrow!

alessandrofelder (approved):

PR #504, opened 2025-02-20
Remove main script

### LINKED ISSUE #273: Decide what to do with  `atlas_generation/main_script`
The idea of this was to have a single script that can run all the atlas generation. We should either:
- Update it to generate all the atlases
- Delete it

cc @viktorpm @alessandrofelder

adamltyson:
This PR removes the main script that was originally intended to allow us to build all atlases and upload them in one go. I don't think this is practical any more considering the number of atlases and the general flakiness of GIN.

I think it's a concept we should come back to eventually, but it's more likely to be an HPC-based batch script rather than a python script that should live here. Happy to be corrected though.

Closes #273


alessandrofelder (approved):
> it's more likely to be an HPC-based batch script rather than a python script that should live here. 

I agree.

Issue #510, opened 2025-03-03
[Feature] add contribution guideline with python env and unit test workflow


mxochicale:
**Is your feature request related to a problem? Please describe.**

Currently there is no contribution guideline nor python env to do dev work

mxochicale:
ERROR using new uv env
```
pip install -e .
Defaulting to user installation because normal site-packages is not writeable
Looking in indexes: https://pypi.org/simple, https://pypi.ngc.nvidia.com
Obtaining file:///home/mxochicale/Desktop/ucl/brainGlobe/repositories/brainglobe-atlasapi
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
ERROR: Project file:///home/mxochicale/Desktop/ucl/brainGlobe/repositories/brainglobe-atlasapi has a 'pyproject.toml' and its build backend is missing the 'build_editable' hook. Since it does not have a 'setup.py' nor a 'setup.cfg', it cannot be installed in editable mode. Consider using a build backend that supports PEP 660.
```

stellaprins:
This may have to do with the setuptools version (currently the requirements are set to <45).

See [this comment](https://stackoverflow.com/questions/69711606/how-to-install-a-package-using-pip-in-editable-mode-with-pyproject-toml) on StackOverflow:
> Note: To be able to do an editable installation to your user site (pip install -e --user), you need a system installed setuptools [v62.0.0 or newer](https://github.com/pypa/setuptools/issues/3019#issuecomment-1264613828).

adamltyson:
The requirements are set to `setuptools>=45`, this seems to work fine with pip on my system (in a conda env).

PR #516, opened 2025-03-04
update setuptools dependency

stellaprins:
## Description

**What is this PR**

- [ ] Bug fix
- [x] Other (update setuptools dependency)

**Why is this PR needed?**
`setuptools` [`v64`](https://setuptools.pypa.io/en/latest/history.html#v64-0-0) introduced editable installs, which is why this version it is often kept as a lower bound in python package dependencies. Not having the `v64` lower bound set often won't give problems since dependency resolvers will usually fetch the more recent versions, but it did in the case of @mxochicale who had an older version of setuptools installed (I believe `v62`).

**What does this PR do?**
Increase lowest `setuptools` version requirement in the `pyproject.toml` to `v64` when editable installs (PEP 660) are introduced.

## References
see [this ](https://github.com/brainglobe/brainglobe-atlasapi/issues/510#issuecomment-2694392420)comment in issue #510 

I changed `setuptools>=45` in `pyproject.toml` to `setuptools==45`, `setuptools==63` and `setuptools==64` etc, and indeed the editable installations is only intstalled, from `v64` onwards.

## Is this a breaking change?
No.

No.

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,7 +71,7 @@ atlasgen = [
 brainglobe = "brainglobe_atlasapi.cli:bg_cli"
 
 [build-system]
-requires = ["setuptools>=45", "wheel", "setuptools_scm[toml]>=6.2"]
+requires = ["setuptools>=64", "wheel", "setuptools_scm[toml]>=6.2"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools]


mxochicale:
Thanks Stella, LGTM, we use uv to accelerate package installation and in this case need to do this `uv pip install --upgrade pip setuptools` to cover setuptools version. Maybe conda users don't need it but feel it is good to keep updating versions.

adamltyson:
Thanks Stella. Not your job of course, but presumably this is something we should eventually change on all our repos?

stellaprins:
> presumably this is something we should eventually change on all our repos?
Yes, that's not a bad idea I think. If editable installs are used it would prevent users having installation issues when for whatever reason their dependency manager installs `setuptools>=63` (probably most likely because that was the version that was found in their base environment).

adamltyson:
Great. I've raised an issue in our central (repo-agnostic) tracker: https://github.com/brainglobe/BrainGlobe/issues/85

adamltyson (approved):

Issue #524, opened 2025-03-10
Validate atlas name


adamltyson:
See https://github.com/brainglobe/brainglobe-atlasapi/pull/523 for context

**To do**
- [x] add more stringent atlas name validation to `validate_atlas_name` in `brainglobe_atlasapi\atlas_generation\validate_atlases.py`.
- [x] adjust `repr` method which currently allows more flexible atlas names than what should be allowed.
- [x] add test with new constraints to `tests\atlasgen\test_validation.py`

adamltyson:
> However after talking to @alessandrofelder and checking the atlas names on [last_versions.conf (GIN)](https://gin.g-node.org/BrainGlobe/atlases/src/master/last_versions.conf) -- they all end with the resolution -- we decided it could make more sense to always expect atlas names to end with the resolution and validate this (if this is not already happening).

Agree. The whole point of the Atlas API is standardisation, so we should validate/enforce it where we can.

stellaprins:
**Naming Convention**

**_allowed name structures:_**
- `[source_id]_[species]_[resolution]`
- `[source_id]_[species]_[additional_info]_[resolution]`

**_allowed characters:_**
- Lowercase letters (a-z)
- Uppercase letters (A-Z)
- Digits (0-9)
- Underscores (_), hyphens (-) and periods (.)

**_Elements in atlas name:_**
- **[source_id]** e.g., `allen`, `kim`, `osten`
- **[species]** e.g., `mouse`, `zfish`, `human`
- **[additional_info]** some atlas names include details about the developmental stage or specific imaging technique used (e.g., `e11-5`, `mri-MTR`, `lsfm`)
- **[resolution]** one or more digits, optionally followed by a period and one or more digits, and ending with "um". (e.g., `37.5um`, `100um`)


Deduced from looking at the atlas names here https://gin.g-node.org/BrainGlobe/atlases/src/master/last_versions.conf

stellaprins:
Just found out there already is a `validate_atlas_name` function which is validates that atlas names do not contain upper case letters. Will add the additional constraints I summarised above under naming convententions.

```
def validate_atlas_name(atlas: BrainGlobeAtlas):
    """
    Ensures atlas names are all lowercase
    """
    assert (
        atlas.atlas_name == atlas.atlas_name.lower()
    ), f"Atlas name {atlas.atlas_name} cannot contain capitals."
    return True
```

adamltyson:
[...]

PR #529, opened 2025-03-12
Increase atlas name validation stringency

stellaprins:
## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**
To validate/enforce stricter naming convention. 

**What does this PR do?**
`validate_atlas_name` checks the following constraints.

Atlas name
- Can contain: lowercase letters, digits, underscores, hyphens, periods
- Ends with resolution (e.g., 5um, 37.5um, 1mm)
- Resolution: number + um (micrometers) or mm (millimeters)

## References
#524

`test_validate_atlas_name` covers:

- **Upper Case Letters**: Names must not contain capital letters. Example: `"CONTAINS_CAPITALS_1um"` should fail.
- **Invalid Characters**: Names must only contain lowercase letters, digits, underscores, hyphens, and periods. Example: `"contains_inv@lid_character_1um"` should fail.
- **Resolution Format**: Names must end with a valid resolution (e.g., `5um`, `1.5mm`). Examples: `"10um_atlas_name_does_not_end_with_resolution"` and `"atlas_name_with_invalid_resolution_100m"` should fail.
- **Valid Names**: Names that meet all criteria should pass. Examples: `"valid_name_100um"` and `"valid-name_1mm"` should pass.

## Is this a breaking change?
No.

No.

## Checklist:

- [x] The documentation has been updated to reflect any changes

### DIFF
--- a/brainglobe_atlasapi/atlas_generation/validate_atlases.py
+++ b/brainglobe_atlasapi/atlas_generation/validate_atlases.py
@@ -249,18 +249,17 @@ def validate_atlas_name(atlas: BrainGlobeAtlas):
     allowed_chars = r"^[a-z0-9_.-]+$"
     res = name.split("_").pop()
 
-    assert (
-        name == name.lower()
-    ), f"Atlas name {atlas.atlas_name} cannot contain capitals."
+    assert name == name.lower(), f"Atlas name {name} cannot contain capitals."
 
     assert re.match(
         allowed_chars, name
     ), f"Atlas name {name} contains invalid characters."
 
     resolution_pattern = r"\d+(\.\d+)?(um|mm)$"
-    assert re.search(
-        resolution_pattern, res
-    ), "Atlas name must end with its resolution (e.g. 5um, 37.5um, 1mm)."
+    assert re.search(resolution_pattern, res), (
+        f"Atlas name {name} should end with a valid resolution "
+        "(e.g., 5um, 1.5mm)."
+    )
 
     return True
 
--- a/brainglobe_atlasapi/atlas_generation/validate_atlases.py
+++ b/brainglobe_atlasapi/atlas_generation/validate_atlases.py
@@ -2,6 +2,7 @@
 
 import json
 import os
+import re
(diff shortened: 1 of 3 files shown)


brainglobe_atlasapi/atlas_generation/validate_atlases.py
```diff
@@ -240,12 +240,27 @@ def validate_annotation_symmetry(atlas: BrainGlobeAtlas):
...
+    res = name.split("_").pop()
+
+    assert name == name.lower(), f"Atlas name {name} cannot contain capitals."
+
+    assert re.match(
+        allowed_chars, name
+    ), f"Atlas name {name} contains invalid characters."
+
+    resolution_pattern = r"\d+(\.\d+)?(um|mm)$"
+    assert re.search(resolution_pattern, res), (
```
adamltyson, line 259: Re https://github.com/brainglobe/brainglobe-atlasapi/issues/524#issuecomment-2724195424, we should decide if we will accept a period or not.

I also wonder if this is a good time to also accept nm? In theory someone could make an EM atlas of a tiny insect.

adamltyson (approved):
Looks good, pending questions about the allowed resolution format.

brainglobe_atlasapi/atlas_generation/validate_atlases.py
stellaprins: I've added `nm`. Although these are now allowed resolutions by the validation, in several other places there is reliance on the resolution ending with `um`. ~~I should have probably addressed this~~ I will address this in the current PR. 

If we want to remove periods from the resolution sections in the atlas names, maybe I can do this in a seperate PR after merging this one.

stellaprins: In theory `atlasapi` can now deal with atlasses that have `nm`, `um`, and `mm` resolutions. In practice however, `um` is still assumed during the atlas generation so only `um` atlasses can be generated.

Issue #514, opened 2025-03-04
Improve test coverage


mxochicale:
codecov for [brainglobe/brainglobe-atlasapi](https://app.codecov.io/gh/brainglobe/brainglobe-atlasapi) is showing 59.19% whereas coverage report for: platform linux, python 3.12.6-final-0 shows 17% (see logs below).
```
#pytest -vs tests/

---------- coverage: platform linux, python 3.12.6-final-0 -----------
Name                                                                                  Stmts   Miss  Cover
---------------------------------------------------------------------------------------------------------
brainglobe_atlasapi/__init__.py                                                           9      2    78%
brainglobe_atlasapi/atlas_generation/__init__.py                                          1      0   100%
brainglobe_atlasapi/atlas_generation/annotation_utils.py                                 36     36     0%
brainglobe_atlasapi/atlas_generation/atlas_scripts/__init__.py                            0      0   100%
brainglobe_atlasapi/atlas_generation/atlas_scripts/admba_3d_dev_mouse.py                152    152     0%
brainglobe_atlasapi/atlas_generation/atlas_scripts/allen_cord.py                        130    130     0%
brainglobe_atlasapi/atlas_generation/atlas_scripts/allen_human.py                       135    135     0%
brainglobe_atlasapi/atlas_generation/atlas_scripts/allen_mouse.py                        47     47     0%
brainglobe_atlasapi/atlas_generation/atlas_scripts/allen_mouse_bluebrain_barrels.py     152    152     0%
brainglobe_atlasapi/atlas_generation/atlas_scripts/australian_mouse.py                  187    187     0%
brainglobe_atlasapi/atlas_generation/atlas_scripts/azba_zfish.py                         86     86     0%
brainglobe_atlasapi/atlas_generation/atlas_scripts/csl_cat.py                           142    142     0%
brainglobe_atlasapi/atlas_generation/atlas_scripts/eurasian_blackcap.py                  99     99     0%
brainglobe_atlasapi/atlas_generation/atlas_scripts/example_mouse.py                      69     69     0%
brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_dev_mouse.py                     173    173     0%
brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_dev_mouse_group.py               121    121     0%
brainglobe_atlasapi/atlas_generation/atlas_scripts/kim_mouse.py                          96     96     0%
[...]

PR #535, opened 2025-03-19
Improve core test coverage

stellaprins:
## Description

**What is this PR**

- [ ] Bug fix
- [x] Other (add tests)

**Why is this PR needed?**
To make code in `brainglobe-atlasapi` more robust by increasing test coverage.

**What does this PR do?**
Adds test for uncovered method `get_structure_mask` in `core`.

## References
#514 

## Is this a breaking change?
No.

No.

## Checklist:

- [x] The documentation has been updated to reflect any changes

### DIFF
--- a/tests/atlasapi/test_core_atlas.py
+++ b/tests/atlasapi/test_core_atlas.py
@@ -13,7 +13,7 @@ def test_initialization(atlas):
     assert atlas.metadata == {
         "name": "example_mouse",
         "citation": (
-            "Wang et al 2020, " "https://doi.org/10.1016/j.cell.2020.04.007"
+            "Wang et al 2020, https://doi.org/10.1016/j.cell.2020.04.007"
         ),
         "atlas_link": "http://www.brain-map.org",
         "species": "Mus musculus",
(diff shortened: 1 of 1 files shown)


tests/atlasapi/test_core_atlas.py
```diff
@@ -187,3 +187,31 @@ def test_even_hemisphere_size(atlas):
...
+def test_get_structure_mask(atlas):
+    """Test the get_structure_mask method.
+
+    Structures in atlas fixture:
+     root (997)
+       └── grey (8)
+             └── CH (567)
+
+    Because the structures (8; 567) are not present in the atlas annotation,
+    they are replaced with 7 and 566.
```
adamltyson, line 201: Could you explain this a bit more? Is this referring to the hierarchy, i.e. the smaller regions contained within the larger region?

stellaprins:
> Would it be possible to add a few more comments about these tests? They look technically fine, but I don't immediately understand what exactly they are doing and why. While I could be less lazy and look into a bit more, I assume that if they confuse me, they will also confuse a new contributor.

It took me a while to figure this out so it would be good to sanity check whether what I have done makes sense. I've added clarification to the test docstring and hope that the example below further clarifies why I reassign the structure ids. 

``` python
from brainglobe_atlasapi.bg_atlas import BrainGlobeAtlas
atlas = BrainGlobeAtlas("example_mouse_100um")
print(atlas.structures)
non_zero_locations = np.nonzero(atlas.annotation)
non_zero_values = atlas.annotation[non_zero_locations]
labels_in_annotation = np.unique(non_zero_values)
print(f"grey (8) present in annotation: {8 in labels_in_annotation}")
print(f"CH (567) present in annotation: {567 in labels_in_annotation}\n")
print(f"7 present in annotation: {7 in labels_in_annotation}")
print(f"566 present in annotation: {566 in labels_in_annotation}")
```
out:
> root (997)
> └── grey (8)
>     └── CH (567)
> 
> grey (8) present in annotation: **False**
> CH (567) present in annotation: **False**
> 
> 7 present in annotation: **True**
> 566 present in annotation: **True**

tests/atlasapi/test_core_atlas.py
stellaprins: It does not refer to the hierarchy, but yes, you're right, it is a bit confusing and looking at this again, maybe I should have reassigned the label numbers in the annotation to match the structure ids instead of reassigning the structure ids? I've tried to make the docstring a bit clearer, it now reads.  Let me know what you think!

>     Test the get_structure_mask method.
> 
>     >>> atlas.structures
>     root (997)
>       └── grey (8)
>             └── CH (567)
> 
>     The 'structures' "grey" and "CH" are present in the example atlas. Their
>     respective ids are 8 and 567. These labels are not present in the
>     annotation of the example atlas however. Because the labels 7 and 566
>     are present, we reassign the parent and substructure ids to match the
>     annotation for testing purposes.
> 
>     Because the "CH" structure is a sub-structure of "grey" it should adopt
>     the parent structure id (7) in the mask where its label (566) is present
>     when get_structure_mask is applied.
> 
>     After applying get_structure_mask only the parent structure id (7) should
>     remain in the mask for the regions corresponding to "CH" and "grey".
> 
>     All labels belonging to structures that are outside of the parent structure
>     should be set to 0.

adamltyson (approved):
Yep, makes sense, thanks!

PR #536, opened 2025-03-19
[ENH] Fix internet connection check for global accessibility

### LINKED ISSUE #522: Check internet connection in countries without Google access.
As found by @NicoKiaru in [an image.sc issue](https://forum.image.sc/t/spinal-cord-atlas-in-abba/108818/12), we ping google to check that there is a network connection. As google is not available in all countries we should probably check a few different sites. I'm not sure what the equivalent in .e.g. China would be.

samadpls:
## Description  

**What is this PR**  
- [x] Bug fix  
- [ ] Other  

**Why is this PR needed?**  
The internet check relied on Google, which is inaccessible in some regions. This PR adds fallback URLs for global accessibility.  

**What does this PR do?**  
Adds alternative URLs (`pypi.org`, `bing.com`, `gitee.com`, `perdu.com`) to ensure connectivity checks work everywhere.  

## References  

Fixes #522  

- Tested locally in environments with and without Google access.  
- Verified fallback URLs work as intended.  

## Is this a breaking change?  
No, it maintains backward compatibility.  

No, functionality remains unchanged for users.  

## Checklist:  
- [ ] Tests have been added  
- [ ] Documentation updated if needed  
- [x] Code formatted with [[pre-commit](https://pre-commit.com/)](https://pre-commit.com/)

### DIFF
--- a/brainglobe_atlasapi/utils.py
+++ b/brainglobe_atlasapi/utils.py
@@ -122,19 +122,33 @@ def check_internet_connection(
     raise_error : bool
         if false, warning but no error.
     """
+    urls_to_try = [url]
 
-    try:
-        _ = requests.get(url, timeout=timeout)
+    if url == "http://www.google.com/":
+        # fallback URLs that are globally accessible
+        fallback_urls = [
+            "https://pypi.org/",
+            "https://www.bing.com/",
+            "https://gitee.com/",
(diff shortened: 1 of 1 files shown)


adamltyson (approved):
Thanks @samadpls!

Issue #562, opened 2025-04-15
[Feature] Atlas API v1.5


IgorTatarnikov:
> We should start to turn this general idea into smaller, actionable issues. Otherwise this will never get done. One idea to start with a version "1.5":
> 
> - Store a copy of metadata outside the packaged atlas
> 
> - Update API to read metadata without downloading atlases
> 
> - Update packaging scripts to package the data without creating an archive
> 
> - Upload files to a new archive
> 
> - Update API to download individual files
> 
> 
> 
> Then we can start adding new features of V2.  

 _Originally posted by @adamltyson in [#141](https://github.com/brainglobe/brainglobe-atlasapi/issues/141#issuecomment-2383192419)_

Just wanted to start this as a separate issue to break this into a smaller issue that can be discussed separately.

I propose that the annotations and templates are stored as `zarr` arrays with fairly standard compression rates. This allows us to keep the download size similar to the current .tar.gz archives, but locally each annotation and template will be stored in a reusable manner that's independent of which atlas requested the data originally.

The other major change that we've discussed is storing the annotations with a `BrainGlobeID` and a mapping to the original annotation IDs to optimise file size on disk, and in memory, but I think this can be pushed toward true V2.

See below for an example directory structure:
```
├── allen_mouse_10um_v1.2.json
├── allen_mouse_25um_v1.2.json
├── annotations_allen_mouse_v3_2017
│   ├── 100um.zarr
│   │   ├── c
│   │   └── zarr.json
│   ├── 10um.zarr
│   │   ├── c
│   │   └── zarr.json
│   ├── 25um.zarr
│   │   ├── c
│   │   └── zarr.json
│   ├── 50um.zarr
│   │   ├── c
│   │   └── zarr.json
│   ├── structures.csv
│   └── structures.json
├── last_versions.conf
├── meshes_allen_mouse_v3_2017
│   ├── 1000.obj
│   ├── 1002.obj
│   ├── 1004.obj
│   ├── 1005.obj
│   ├── 1006.obj
│   ├── 1007.obj
│   ├── 1008.obj
│   ├── etc...
│  
└── reference_allen_mouse_v1
    ├── 100um.zarr
    │   ├── c
    │   └── zarr.json
    ├── 10um.zarr
    │   ├── c
    │   └── zarr.json
    ├── 25um.zarr
    │   ├── c
    │   └── zarr.json
    └── 50um.zarr
        ├── c
        └── zarr.json
``` 

[...]

PR #563, opened 2025-04-24
Initial V1.5 implementation

IgorTatarnikov:

## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**
See #562 

**What does this PR do?**
This PR creates a new `BrainGlobeAtlasV2` class that inherits from `BrainGlobeAtlas` and overrides any methods that need to be changed for V1.5 to work.

Downloading an atlas to a fresh `.brainglobe` directory triggers the following:
- Local name for the requested atlas is not found (this specifically checks for the `json` e.g. `allen_mouse_100um_v1.2.json`
- The `json` is fetched from the [GIN](https://gin.g-node.org/BrainGlobe/atlases-v2) repository
- The metadata from the `json` is read, and directories for the annotations, template, and meshes are created
- `structures.json` and `structures.csv` are fetched for the appropriate annotations
- The `BrainGlobeAtlasV2` object is instantiated

The actual annotations and template images are not fetched at this time, nor are the meshes.

When the user requests the annotations or references by calling `atlas.reference` or `atlas.annotation` the respective data is fetched at the correct resolution. Meshes are fetched individually when they are requested for now.

Things that still need attention/discussion:
- Need to add a function/API endpoint to force download all data for a given atlas
- Fetch the meshes in a grouped way some how (parent, and 2 levels of children perhaps?)
- Print/log a warning the first time the annotation/template data is requested to inform the user it has to be downloaded first
- I'm also not a fan of needing to download the `structures.json` and `structures.csv` to instantiate the Atlas object but that would involve more changes to `core.Atlas` which I've tried to avoid for now.
- The function that sets up the atlas is named `download_extract_file` even though we're not longer extracting, this is to keep compatibility with the `__init__` method of `BrainGlobeAtlas`.

## References
#562

Added tests that compare the data between a V1 atlas and a V2 atlas where possible.

## Is this a breaking change?
Sort of. All V2 atlases live in a `.brainglobe/atlases` directory, ensuring compatibility with the current V1 layout. However, the cached `last_versions.conf` file is overwritten in the `.brainglobe` directory.

Will have to come later.

## Checklist:

### DIFF
--- a/brainglobe_atlasapi/bg_atlasv2.py
+++ b/brainglobe_atlasapi/bg_atlasv2.py
@@ -1,6 +1,13 @@
+import tarfile
+from pathlib import Path
+
 import zarr
 
-from brainglobe_atlasapi import BrainGlobeAtlas, descriptors
+from brainglobe_atlasapi import BrainGlobeAtlas, descriptors, utils
+from brainglobe_atlasapi.utils import (
+    check_gin_status,
+    check_internet_connection,
+)
 
 
 class BrainGlobeAtlasV2(BrainGlobeAtlas):
@@ -47,9 +54,80 @@ class BrainGlobeAtlasV2(BrainGlobeAtlas):
                 / self.metadata["reference_images"][0]
                 / f"{int(self.resolution[0])}um"
             )
-            zarr_array = zarr.open_array(
+            if not reference_path.exists():
+                remote_url = self._remote_url_base.format(
+                    f"{self.metadata['reference_images'][0]}/{int(self.resolution[0])}um.tar.gz"
+                )
+                self.download_tar_file(remote_url, reference_path)
+
+            reference_array = zarr.open_array(
                 reference_path, mode="r", zarr_format=3
             )
-            self._reference = zarr_array[:]
+            self._reference = reference_array[:]
 
         return self._reference
+
+    @property
+    def annotation(self):
+        if self._annotation is None:
+            annotation_path = (
+                self.root_dir
+                / "annotations"
+                / self.metadata["annotation_images"][0]
+                / f"{int(self.resolution[0])}um"
+            )
+            if not annotation_path.exists():
+                remote_url = self._remote_url_base.format(
+                    f"{self.metadata['annotation_images'][0]}/{int(self.resolution[0])}um.tar.gz"
+                )
+                self.download_tar_file(remote_url, annotation_path)
+
+            annotation_array = zarr.open_array(
+                annotation_path, mode="r", zarr_format=3
+            )
+            self._annotation = annotation_array[:]
+
+        return self._annotation
+
+    def download_tar_file(self, url: str, local_path: Path):
+        """Download and extract a file from a URL."""
+        # Implement the download and extraction logic here
+        destination_path = self.interm_download_dir / local_path.with_suffix(
+            ".tar.gz"
+        )
+        utils.retrieve_over_http(url, destination_path, self.fn_update)
+
+        # Uncompress the downloaded file
+        tar = tarfile.open(destination_path)
+        tar.extractall(path=local_path.parent)
+        tar.close()
+
+        destination_path.unlink()
+
+    @property
+    def remote_url(self):
+        if self.remote_version is not None:
+            name = (
+                f"{self.atlas_name}_v{self.remote_version[0]}."
+                f"{self.remote_version[1]}.json"
+            )
+
+            return self._remote_url_base.format(name)
+
+    def download_extract_file(self):
+        """Download and extract atlas from remote url."""
+        check_internet_connection()
+        check_gin_status()
+
+        name = (
+            f"{self.atlas_name}_v{self.remote_version[0]}."
+            f"{self.remote_version[1]}.json"
+        )
+
+        # Get path to folder where data will be saved
+        destination_path = self.brainglobe_dir / name
+
+        # Try to download atlas data
+        utils.retrieve_over_http(
+            self.remote_url, destination_path, self.fn_update
+        )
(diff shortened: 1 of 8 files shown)


tests/conftest.py
```diff
@@ -51,6 +51,21 @@ def mock_home():
         monkeypatch.setattr(config, "TEMPLATE_CONF_DICT", mock_default_dirs)
 
 
+@pytest.fixture(autouse=True)
+def setup_preexisting_local_atlases():
+    """Automatically setup all tests to have three downloaded atlases
```
adamltyson, line 56: Should this be:
```suggestion
    """Automatically setup all tests to have two downloaded atlases
```

brainglobe_atlasapi/bg_atlasv2.py
```diff
@@ -0,0 +1,259 @@
...
+    def template(self):
+        if self._template is None:
+            template_path = (
+                self.root_dir
+                / "templates"
+                / self.metadata["reference_images"][0]
+                / f"{int(self.resolution[0])}um"
+            )
+            if not template_path.exists():
+                remote_url = self._remote_url_base.format(
```
adamltyson, line 72: Not needed now, but at some point I think the different types of data (template, annotation etc) should be in their own directory on GIN. Each subdir can then have a simpler name too.

brainglobe_atlasapi/bg_atlasv2.py
```diff
@@ -0,0 +1,259 @@
...
+        return self._annotation
+
+    def download_tar_file(self, url: str, local_path: Path):
+        """Download and extract a file from a URL."""
+        # Implement the download and extraction logic here
+        destination_path = (
+            self.interm_download_dir / local_path.with_suffix(".tar.gz").name
+        )
+
+        utils.retrieve_over_http(url, destination_path, self.fn_update)
```
adamltyson, line 118: Can pooch be used rather than our custom function?

brainglobe_atlasapi/bg_atlasv2.py
```diff
@@ -0,0 +1,259 @@
...
+        """
+        version_str = self.metadata["version"]
+
+        if version_str is None:
+            return None
+
+        return tuple(int(version) for version in version_str.split("."))
+
+    @property
+    def template(self):
```
adamltyson, line 63: We should come up with a more sophisticated solution at some point, but there should be something printed here along the lines of "Downloading template image..."

brainglobe_atlasapi/bg_atlasv2.py
```diff
@@ -0,0 +1,259 @@
...
+        template_dir = (
+            destination_path.parent
+            / "templates"
+            / self.metadata["reference_images"][0]
+        )
+        meshes_dir = (
+            destination_path.parent / "meshes" / self.metadata["meshes"][0]
+        )
+
+        if not annotations_dir.exists():
```
adamltyson, line 175: Are the conditional statements here needed if you can use `exist_ok=True`?

brainglobe_atlasapi/core.py
```diff
@@ -31,18 +31,33 @@ class Atlas:
...
-        self.metadata = read_json(self.root_dir / METADATA_FILENAME)
+        atlas_path = Path(path)
+        # v1
+        if atlas_path.is_dir():
+            self.root_dir = atlas_path
+            self.metadata = read_json(self.root_dir / METADATA_FILENAME)
+            structures_list = read_json(self.root_dir / STRUCTURES_FILENAME)
+            meshes_dir = MESHES_DIRNAME
+        # v2
+        elif atlas_path.suffix == ".json":
```
adamltyson, line 42: To simplify things in the future, could we move to yaml now? Unless there's a specific reason not to?

adamltyson (approved):
Looks good, and seems to work locally flawlessly!

I have a million comments about V2 generally, but for the sake of keeping some momentum, I've tried to restrict my comments to things specifically about this PR.

:rocket:

brainglobe_atlasapi/core.py
IgorTatarnikov: No specific reason, I was just piggy backing off of the `read_json` function. I'll change to yaml and update the GIN repository.

brainglobe_atlasapi/bg_atlasv2.py
IgorTatarnikov: :+1:

brainglobe_atlasapi/bg_atlasv2.py
IgorTatarnikov: I can try, I wasn't sure where to store the `known_hashes` for each component of the atlas. This could be in a registry file that's stored on GIN, or as part of the metadata in the atlas? I can just use `None` for now to replace the logic with `pooch`.

adamltyson: Yeah I think it can just be `None` for now

brainglobe_atlasapi/bg_atlasv2.py
IgorTatarnikov: In order to get pooch to work nicely, it helped for the remote directory structure to match the local one. I've reorganised the files on GIN, and simplified the subdir names.

brainglobe_atlasapi/core.py
IgorTatarnikov: Done, atlases are now stored as yaml files

IgorTatarnikov:
The data is now fetched using `pooch` and atlases are stored as yaml files. @adamltyson did you want to take another look?

brainglobe_atlasapi/utils.py
```diff
@@ -247,6 +248,42 @@ def retrieve_over_http(
         )
 
 
+def retrieve_over_http_pooch(
```
adamltyson, line 251: I'm probably missing something, but where is this function used?

IgorTatarnikov: That's a remnant of the first approach I tried, I'll remove it

PR #565, opened 2025-04-30
Use correct directory for downloading atlases on CI

### LINKED ISSUE #564: [BUG] CI cache issues
Running the test suite locally takes ~2 minutes, whereas it takes ~30 minutes on CI.

I noticed that the tests pause any time a new atlas is required to be loaded suggesting the cache isn't being saved/retrieved correctly.

IgorTatarnikov:
I found the issue: https://github.com/brainglobe/brainglobe-atlasapi/blob/638091910cc13ca8093a153beb3dd57f28c714b6/tests/conftest.py#L24-L28 The conditional statement was always true even on CI because the `GITHUB_ACTIONS` env variable wasn't being passed to `tox`. Since we only cache the `.brainglobe` directory it was downloading fresh atlases into a `.brainglobe-tests` directory on each run.

The env variable is now being passed back and the cache is being utilised correctly. It now takes ~12 minutes to run the tests. However, this should be sped up when the cache on main is refreshed with new atlases that have been added to the test suite since the last time the cache was used correctly. We can force a refresh by deleting the `brainglobe-{os}-18` caches on main after this is merged and rerunning the test action.

I've also updated one test where a fresh atlas was being downloaded to use the one we have in the cache. It would be nice to go through the testing suite and ensure we're only downloading a new atlas when required. GIN gets grumpy easily and harshly throttles connections leading to long test times.

Closes #564

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -40,12 +40,12 @@ jobs:
       matrix:
         # Run all supported Python versions on linux
         os: [ubuntu-latest]
-        python-version: ["3.11", "3.12", "3.13"]
-        include:
-          - os: macos-latest
-            python-version: "3.13"
-          - os: windows-latest
-            python-version: "3.13"
+        python-version: ["3.13"]
+#        include:
+#          - os: macos-latest
+#            python-version: "3.13"
+#          - os: windows-latest
+#            python-version: "3.13"
 
     steps:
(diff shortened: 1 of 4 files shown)


alessandrofelder (approved):
Excellent detective work and super nice fix AFAICT :100: (I hope we're right and it's sped up once the main cache is fresh :crossed_fingers: :sweat_smile: )

PR #575, opened 2025-05-29
add ccfv3a to table

PolarBean:
Just updates the table. The link to the documentation will be functional once https://github.com/brainglobe/brainglobe.github.io/pull/336 is merged :)


alessandrofelder (approved):
Thanks, @PolarBean - we will need to think more about modality versus stain (#578) but I consider that out-of-scope for this PR for now.

PR #586, opened 2025-06-06
Add additional_references keys on construction

### LINKED ISSUE #583: [BUG] additional reference not showing up in CCFv3a
For some reason when i look at the additional_references for the BBP CCFv3a atlas it returns an empty dictionary. 

Image

This is strange as in the metadata of the atlas file in the .brainglobe directory it does include additional references. 

Image

and the additional reference volume is present in the directory. 

Image

Code to reproduce:
```python
from brainglobe_atlasapi.bg_atlas import BrainGlobeAtlas
atlas = BrainGlobeAtlas("ccfv3augmented_mouse_10um")
print(atlas.additional_references)
```

IgorTatarnikov:

## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**
See #583 

**What does this PR do?**
On `AdditionalRefDict` construction all additional references passed in the `references_list` are added to the dict with `None` as the entry. When `__getitem__` is called, it now checks if `self.data[ref_name] is None` to see if the reference has been loaded to memory, if not it opens the additional reference file and saves it in the dict.

## References
Closes #583 

All existing tests pass.

## Is this a breaking change?
No

Yes. We should document the canonical way to fetch a list of names for the additional references from within the API.

Something like `tuple(atlas.additional_references)`?

## Checklist:

### DIFF
--- a/brainglobe_atlasapi/core.py
+++ b/brainglobe_atlasapi/core.py
@@ -358,14 +358,14 @@ class AdditionalRefDict(UserDict):
             self.data[ref_name] = None
 
     def __getitem__(self, ref_name):
-        if self.data[ref_name] is None:
-            if ref_name not in self.references_list:
-                warnings.warn(
-                    f"No reference named {ref_name} "
-                    f"(available: {self.references_list})"
-                )
-                return None
+        if ref_name not in self.references_list:
+            warnings.warn(
+                f"No reference named {ref_name} "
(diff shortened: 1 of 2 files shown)


IgorTatarnikov:
Yup! This feels like something tha should be immutable.  I don't like the atlernative version, if we want direct access I think it should be a property of `BrainGlobeAtlas` itself.

PR #587, opened 2025-06-17
[New Atlas]: Adding basic bumblebee atlas

Wolfffff:
# [New Atlas]: Adding basic bumblebee atlas

> **Note** This is a _very_ rough PR. I'd appreciate some feedback on both the structure and implementation.

**Notes**
Along the way, I noted a couple of small issues that are **not** fixed in this PR. I can work on separately as required:

- Figure out how to correctly represent resolution.

The real resolution is 0.64µm x 0.64µm x 2.542µm but the mesh additions only work when using 1. Further, the naming expects an integer.

- Missing method in wrapup.py: 
 
The current wrap-up script references an instance method that doesn't exist. Specifically, [wrapup.py line 249](https://github.com/brainglobe/brainglobe-atlasapi/blob/main/brainglobe_atlasapi/atlas_generation/wrapup.py#L249) calls `map_resolution` on an `AnatomicalSpace` object, which doesn’t implement that method ([core.py line 30](https://github.com/brainglobe/brainglobe-space/blob/main/brainglobe_space/core.py#L30)).

- Encoding errors in tree files:  

I encountered encoding errors in [create_readme](https://github.com/brainglobe_atlasapi/brainglobe-atlasapi/blob/main/brainglobe_atlasapi/atlas_generation/metadata_utils.py#L85) -- specifically due to tree characters causing I/O failures. Forcing UTF-8 encoding was a sufficient fix here.

- Atlas naming validation:

The atlas naming validation breaks when following the base template (e.g., kocher_bumblebee), since the name does not include a resolution as required [here](https://github.com/Wolfffff/brainglobe-atlasapi/blob/main/brainglobe_atlasapi/atlas_generation/validate_atlases.py#L259).

## Description

**What is this PR**

- [ ] Bug fix
- [x] Addition of a new feature
- [ ] Other

**Why is this PR needed?**

This PR expands BrainGlobe's atlas collection to include the first arthropod species - the common eastern bumblebee (*Bombus impatiens*). This atlas provides coarse annotations of key neuropil regions in the bumblebee brain.

**What does this PR do?**

This PR adds an atlas generation script for the Kocher bumblebee brain atlas, including:

- Downloads VTK mesh and template data from the Kocher lab
- Converts VTK mesh annotations to volumetric TIFF format
- Generates 3D meshes for visualization of 9 brain regions
- Handles coordinate transformations and padding to align annotations with reference template
- Integrates with existing BrainGlobe Atlas API conventions

## How has this PR been tested?

[...]

PR #594, opened 2025-06-25
[588]: Removing function that doesnt exist, passing resolution directly for now

Wolfffff:
## Description

Removing function that doesn't exist, passing resolution directly for now. 

Warning: This could cause issues if the resolution is passed in an incorrect order. This issue propagates to [brainglobe-registration](https://github.com/brainglobe/brainglobe-registration/blob/a799f834b68ac0c29e99cb6718e2f75a898be8e5/brainglobe_registration/registration_widget.py#L708-L722) now as well since the ASR resolution is out of order in the scaling.

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**

It was calling a method that doesn't exist.

**What does this PR do?**

## References

#587 

Run locally.

## Is this a breaking change?

It could? @alessandrofelder if you could confirm that there are no cases this would break, that'd be awesome.

No.

## Checklist:

### DIFF
--- a/brainglobe_atlasapi/atlas_generation/wrapup.py
+++ b/brainglobe_atlasapi/atlas_generation/wrapup.py
@@ -253,7 +253,7 @@ def wrapup_atlas_from_data(
         atlas_link=atlas_link,
         species=species,
         symmetric=symmetric,
-        resolution=resolution, # We expect input to be asr
+        resolution=resolution,  # We expect input to be asr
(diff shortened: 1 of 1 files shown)


alessandrofelder:
Hmmm, on second thought, it could be a bug on the `brainglobe-registration` side that we haven't notice because we don't test with anisotropic atlases often :thinking:

I'm pretty certain this fix (and the `resolution_mapping` and ASR convention) is what we'd like to have on the atlas packaging side.

PR #607, opened 2025-07-04
A test for write_stack function

Skxsmy:

## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**
Currently we lack a test module for stacks.py.
**What does this PR do?**
Adding the test_stacks.py, with a test function for write_stack function.
## References
\
Locally
## Is this a breaking change?
No
No
## Checklist:

### DIFF
--- a/tests/atlasgen/test_stacks.py
+++ b/tests/atlasgen/test_stacks.py
@@ -9,10 +9,10 @@ from brainglobe_atlasapi.atlas_generation.stacks import write_stack
 
 
 @pytest.fixture
-def image_unit16():
+def image_uint16():
     """Fixture providing a dummy image stack."""
-    image_unit16 = np.random.randint(0, 65535, (10, 10, 10), dtype=np.uint16)
-    yield image_unit16
+    image_uint16 = np.random.randint(0, 65535, (10, 10, 10), dtype=np.uint16)
+    yield image_uint16
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


tests/atlasgen/test_stacks.py
```diff
@@ -0,0 +1,49 @@
...
+
+import numpy as np
+import pytest
+import tifffile
+
+from brainglobe_atlasapi.atlas_generation.stacks import write_stack
+
+
+@pytest.fixture
+def image_unit():
```
alessandrofelder:
```suggestion
def image_uint16():
```
I think? Unless I am missing what "unit" refers to here.
And rename accordingly in the remainder of the file.

tests/atlasgen/test_stacks.py
```diff
@@ -0,0 +1,49 @@
...
+    """Fixture providing a dummy image stack."""
+    image_unit = np.random.randint(0, 65535, (10, 10, 10), dtype=np.uint16)
```
alessandrofelder:
```suggestion
    image_uint16 = np.random.randint(0, 65535, (10, 10, 10), dtype=np.uint16)
```

tests/atlasgen/test_stacks.py
```diff
@@ -0,0 +1,49 @@
...
+    """Fixture providing a dummy image stack."""
+    image_float = np.random.rand(10, 10, 10).astype(np.float32)
+    yield image_float
+
+
+@pytest.fixture
+def tmp_path():
+    """Fixture providing a temporary directory for testing."""
+    with tempfile.TemporaryDirectory() as tmpdir:
+        yield tmpdir
```
alessandrofelder:
```suggestion
```

Pytest already has [a fixture for this](https://docs.pytest.org/en/stable/how-to/tmp_path.html), which you can use directly without redefining it

tests/atlasgen/test_stacks.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def test_write_stack(image_unit, image_float, tmp_path):
```
alessandrofelder: I suggest splitting this into two tests, one for the integer image and one for the floating point image.

tests/atlasgen/test_stacks.py
```diff
@@ -0,0 +1,46 @@
...
+
+import numpy as np
+import pytest
+import tifffile
+
+from brainglobe_atlasapi.atlas_generation.stacks import write_stack
+
+
+@pytest.fixture
+def image_unit16():
```
alessandrofelder:
```suggestion
def image_uint16():
```
Sorry, I should have been clearer in my previous review... think there is still some confusion around this and related variable names?

I think the variable names should all end in "uint16" (not "unit16") - this refers to the pixel values in the test imaging being unsigned 16-bit integers, or uint16 for short.

alessandrofelder (changes requested):
Small bit of confusion in variables names to clear up, then we are ready to merge 🎉

tests/atlasgen/test_stacks.py
Skxsmy: I see, have corrected them!

PR #609, opened 2025-07-07
use unit consistently across conditionals in `atlas_name_from_repr`

alessandrofelder:
## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**

A small bug is blocking #587 : `atlas_name_from_repr` it _should_ always return a unit in the atlas name representation, but it currently only does so only if you also specify major and minor version.

**What does this PR do?**

Tiny bugfix in utility function used by wrapup: `atlas_name_from_repr` always returns unit irrespective of whether minor and major versions are specified or not, making the behaviour consistent.

### DIFF
--- a/brainglobe_atlasapi/utils.py
+++ b/brainglobe_atlasapi/utils.py
@@ -123,7 +123,7 @@ def atlas_name_from_repr(
 ):
     """Generate atlas name given a description."""
     if major_vers is None and minor_vers is None:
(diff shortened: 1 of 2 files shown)


IgorTatarnikov (approved):
Looks good to me! Tested the functions separately, and also tested packaging the black cap atlas (I ran into this bug when packaging and it's solved now!)

PR #628, opened 2025-08-21
Spell out acronyms and reformat non isotropic voxel size in readme

PolarBean:
Very nitpicky I just wanted to make sure we were consistent in the way we describe non isotropic voxel sizes. This just makes the kocher bumble bee description consistent with the allen spinal cord.

Additionally I propose spelling out each acronym the first time its used.


adamltyson:
Thanks for this @PolarBean. I disagree with spelling out the acryonym just the first time though. As it's a table (and not prose), readers may not go through it all in order. I think we should either spell it out on every line, or add some kind of legend/caption at the bottom to explain acronyms. I prefer the second to save space.

PolarBean:
I agree that's nicer i'll put something together

adamltyson:
Looks good, I just noticed two things:
- The Max Planck zfish atlas no longer has an age, just "dpf"
- The Gubra stereotaxic atlas age has some unnecessary hyphens (should be the same formatting as the Australian Mouse Atlas) 

Would you mind making a similar PR to the website?

PR #636, opened 2025-09-01
Drosophila wingdisc v1

alessandrofelder:
Packages v1 of the drosophila wing disc atlas, as the version submitted as part of @Skxsmy's thesis. cc @giuliapaci

Supercedes #616


adamltyson (approved):
LGTM. I haven't tested this, but I assume it's been run successfully considering the atlas is in the API!

I will let you merge it @alessandrofelder just in case this does require a deeper review, but I'm happy with it.

alessandrofelder:
The version I uploaded yesterday was made with this, so I expect it "works". Thanks for looking at the code changes!

PR #681, opened 2025-11-28
Fix #668: Safely handle empty custom_atlases.conf file

### LINKED ISSUE #668: [BUG] Empty custom_atlases.conf file leads to crash
If the `custom_atlases.conf` file is present in the BrainGlobe directory but empty `brainglobe list` results in a crash.

Traceback below:

```
Traceback (most recent call last):
  File "/home/igor/miniforge3/envs/brainglobe-atlasapi/bin/brainglobe", line 8, in <module>
    sys.exit(bg_cli())
             ^^^^^^^^
  File "/home/igor/miniforge3/envs/brainglobe-atlasapi/lib/python3.12/site-packages/click/core.py", line 1442, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/igor/miniforge3/envs/brainglobe-atlasapi/lib/python3.12/site-packages/click/core.py", line 1363, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/igor/miniforge3/envs/brainglobe-atlasapi/lib/python3.12/site-packages/click/core.py", line 1226, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/igor/miniforge3/envs/brainglobe-atlasapi/lib/python3.12/site-packages/click/core.py", line 794, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/igor/NIU-dev/bg-atlasapi/brainglobe_atlasapi/cli.py", line 43, in bg_cli
    return show_atlases(show_local_path=show)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/igor/NIU-dev/bg-atlasapi/brainglobe_atlasapi/list_atlases.py", line 124, in show_atlases
    available_atlases = get_all_atlases_lastversions()
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/igor/NIU-dev/bg-atlasapi/brainglobe_atlasapi/list_atlases.py", line 76, in get_all_atlases_lastversions
    return {**official_atlases["atlases"], **custom_atlases["atlases"]}
                                             ~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/home/igor/miniforge3/envs/brainglobe-atlasapi/lib/python3.12/configparser.py", line 941, in __getitem__
    raise KeyError(key)
KeyError: 'atlases'
```

kalpana-pardeep-rassani:

## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**
The current implementation crashes with a `KeyError: 'atlases'` if the local `custom_atlases.conf` file exists but is empty (contains no sections). This prevents the `brainglobe list` command from running.

**What does this PR do?**
This PR modifies the `get_all_atlases_lastversions` function in `brainglobe_atlasapi/list_atlases.py` to use a safe `.get("atlases", {})` method when reading the custom atlas configuration. This ensures that an empty dictionary is used if the 'atlases' section is missing, thus preventing the crash.

## References

Fixes #668

The fix was tested locally by ensuring that the `brainglobe list` command runs successfully without crashing when the `~/.brainglobe/custom_atlases.conf` file is present but completely empty.

## Is this a breaking change?

No. This is a fix for an existing bug and does not change any public API or required functionality.

No.

## Checklist:

- [done] The code has been tested locally
- [done] Tests have been added to cover all new functionality (unit & integration)
- [done] The code has been formatted with [pre-commit](https://pre-commit.com/)
Note on Test Failures:

All code changes related to Issue #668 (empty config file) have been verified locally. When running the full test suite, I encountered 2 ConnectionErrors (due to the GIN server being unreachable) and 1 FileNotFoundError (due to missing local test data), which are external to this fix. The patch itself is clean.

### DIFF
--- a/brainglobe_atlasapi/list_atlases.py
+++ b/brainglobe_atlasapi/list_atlases.py
@@ -76,7 +76,8 @@ def get_all_atlases_lastversions():
 
     # --- FIX APPLIED HERE ---
     # Safely get the custom atlases data. If the 'atlases' section does not
-    # exist (e.g., file is empty), .get("atlases", {}) returns {} instead of crashing.
+    # exist (e.g., file is empty), .get("atlases", {}) returns {} instead
+    # of crashing.
     custom_atlas_data = custom_atlases.get("atlases", {})
     return {**official_atlases["atlases"], **custom_atlas_data}
     # ------------------------
--- a/tests/atlasapi/test_list_atlases.py
+++ b/tests/atlasapi/test_list_atlases.py
@@ -1,7 +1,7 @@
 """Test functions for listing and managing BrainGlobe atlases."""
 
-from unittest import mock
 from pathlib import Path
+from unittest import mock
 
 import pytest
 from rich.console import Console
@@ -21,24 +21,23 @@ from brainglobe_atlasapi.list_atlases import (
 @pytest.fixture
 def mock_atlas_dir(tmp_path, mocker):
     """
-    Creates a mock atlas directory and patches get_brainglobe_dir
+    Create a mock atlas directory and patch get_brainglobe_dir
     to point to the temporary location.
     """
     # 1. Create the mock brainglobe directory structure
     bg_dir = tmp_path / ".brainglobe"
     bg_dir.mkdir()
-    
+
     # 2. Create the mock atlas directory with a version tag
     atlas_name_versioned = "example_mouse_100um_v1.0"
     atlas_dir = bg_dir / atlas_name_versioned
     atlas_dir.mkdir()
-    
+
     # 3. Patch the function that returns the brainglobe directory
     mocker.patch(
-        "brainglobe_atlasapi.config.get_brainglobe_dir",
-        return_value=bg_dir
+        "brainglobe_atlasapi.config.get_brainglobe_dir", return_value=bg_dir
     )
-    
+
     # Return the name of the atlas for use in tests
     return "example_mouse_100um"
 
@@ -58,7 +57,8 @@ def test_get_local_atlas_version_real_atlas(mock_atlas_dir):
 
 
 def test_get_local_atlas_version_missing_atlas(capsys):
-    """Test retrieving the version of a non-existent atlas.
+    """
+    Test retrieving the version of a non-existent atlas.
 
     Parameters
     ----------
@@ -73,10 +73,9 @@ def test_get_local_atlas_version_missing_atlas(capsys):
 
 def test_lastversions(mock_atlas_dir, mocker):
     """Test retrieving atlas versions from the online source."""
-    
     # Mock the remote/cached official versions to include our mock atlas
     mock_official_atlases = {"atlases": {mock_atlas_dir: "1.0"}}
-    
+
     # Patch utils.conf_from_url to return our mock data
     mocker.patch(
         "brainglobe_atlasapi.list_atlases.utils.conf_from_url",
@@ -88,9 +87,9 @@ def test_lastversions(mock_atlas_dir, mocker):
         side_effect=lambda file_path: (
             mock_official_atlases
             if file_path.name == "last_versions.conf"
-            # FIX APPLIED HERE: Return {} instead of FileNotFoundError 
-            # to avoid the AttributeError when the test tries to access the result.
-            else {} 
+            # Return {} instead of FileNotFoundError to avoid
+            # the AttributeError when the test tries to access the result.
+            else {}
         ),
     )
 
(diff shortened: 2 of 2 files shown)


brainglobe_atlasapi/list_atlases.py
```diff
@@ -69,11 +70,23 @@ def get_all_atlases_lastversions():
...
+    # --- FIX APPLIED HERE ---
+    # Safely get the custom atlases data. If the 'atlases' section does not
+    # exist (e.g., file is empty), custom_atlases.get("atlases", {}) returns {}
+    # instead of crashing.
+    custom_atlas_data = custom_atlases.get("atlases", {})
+
+    return {**official_atlases["atlases"], **custom_atlas_data}
+
+
+# The get_atlases_lastversions function below this should remain unchanged.
```
alessandrofelder:
```suggestion
    # Safely get the custom atlases data. If the 'atlases' section does not
    # exist (e.g., file is empty), custom_atlases.get("atlases", {}) returns {}
    # instead of crashing.
    custom_atlas_data = custom_atlases.get("atlases", {})

    return {**official_atlases["atlases"], **custom_atlas_data}


```

I think some of these comments relate to the PR, and not the code itself - so I don't think they belong in the code. Do you agree (I may have missed something)?

brainglobe_atlasapi/list_atlases.py
```diff
@@ -228,3 +241,32 @@ def add_atlas_to_row(atlas, info, table, show_local_path=False):
     table.add_row(*row)
 
     return table
+
+
+def test_empty_config_file_no_crash(tmp_path):
```
alessandrofelder: This test looks reasonable at first glance but I think it should go somewhere in the `tests` folder with the other tests. It will be picked up by `pytest` there!

brainglobe_atlasapi/list_atlases.py
```diff
@@ -4,6 +4,7 @@
 """
 
 import re
+import warnings  # Adding this line
```
alessandrofelder:
```suggestion
import warnings
```
I don't think this comment is useful - it is very generic and not related to the line of code AFAICT

brainglobe_atlasapi/list_atlases.py
```diff
@@ -73,7 +73,14 @@ def get_all_atlases_lastversions():
         custom_atlases = utils.conf_from_file(custom_path)
     except FileNotFoundError:
         return dict(official_atlases["atlases"])
-    return {**official_atlases["atlases"], **custom_atlases["atlases"]}
+
+    # --- FIX APPLIED HERE ---
```
alessandrofelder, line 77:
```suggestion
```
This will not mean anything after this issue is fixed - it's too generic and makes the code harder to interpret. So I think it should be remove.

tests/atlasapi/test_list_atlases.py
```diff
@@ -17,22 +18,47 @@
 )
 
 
-def test_get_downloaded_atlases():
+@pytest.fixture
+def mock_atlas_dir(tmp_path, mocker):
```
alessandrofelder, line 22: Most of the functionality in this fixture is already implemented in https://github.com/brainglobe/brainglobe-atlasapi/blob/main/tests/conftest.py and ready for you to use here, so I would suggest
* removing it here
* refactoring your test to use the existing functionality
so we don't have to look after duplicate code.

tests/atlasapi/test_list_atlases.py
```diff
@@ -203,3 +243,52 @@ def test_add_atlas_to_row(version, expected_print, capsys):
     Console().print(table)
     captured = capsys.readouterr()
     assert expected_print in captured.out
+
+
+def test_empty_custom_config_no_crash(tmp_path: Path, mocker):
```
alessandrofelder, line 248: I think it is sufficient to only add this test, and not change any of the other tests - they test things that are unrelated to the issue we are trying to fix.

I also think this test could be a lot simpler. You only need to check that 
* when you have an empty `custom_atlases.conf` file...
*... then the function doesn't crash, and possibly returns some existing atlases that are not custom.

It shouldn't need to be this long, or involve so many mocks. You can re-use some existing fixtures from `conftest.py` as I suggested in a previous comment.

Issue #689, opened 2025-12-03
[Enhancement]  Add missing type annotations across core atlas API modules


nishanthcr7777:
**Is your feature request related to a problem? Please describe.**  

The `brainglobe-atlasapi` codebase currently lacks type annotations in several core files.  
This makes the code harder to read, reduces IDE autocompletion, and limits static analysis with tools like `mypy` (which is already included as a development dependency).  
Adding type hints will improve clarity, maintainability, and contributor onboarding.

---

**Describe the solution you'd like**  

I propose adding standard Python type annotations to the following modules and functions that currently have untyped parameters or return values.

### 1. `brainglobe_atlasapi/core.py`

The `Atlas` class contains multiple untyped methods. Examples:

- `get_structure_descendants(self, structure)`
  - structure: `str | int`
  - return: `list[str]` (returns list of acronyms)

- `get_structure_ancestors(self, structure)`
  - structure: `str | int`
  - return: `list[str]`

- `hemisphere_from_coords(self, coords, microns=False, as_string=False)`
  - coords: `tuple | list | np.ndarray`
  - return: `int | str`

### 2. `brainglobe_atlasapi/utils.py`

- `read_json(path)`
  - path: `str | Path`
  - return: `dict`

- `read_tiff(path)`
  - path: `str | Path`
  - return: `np.ndarray`

### 3. `brainglobe_atlasapi/structure_tree_util.py`

- `child_ids(structure, structure_list)`
  - structure: `int`
    - (Note: current docstring incorrectly states `dict`, but source code expects an integer ID.)
  - structure_list: `list`
  - return: `list[int]`

- `get_structures_tree(structures_list)`
  - structures_list: `list`
  - return: `Tree` (from `treelib`)

---

**Describe alternatives you've considered**  

The only alternative is keeping the current untyped structure, but this reduces the effectiveness of static type checking, increases cognitive load for contributors, and introduces ambiguity in function behavior.  
Type annotations provide clearer documentation with no runtime overhead.

---

**Additional context**  

This proposal is fully non-breaking and only adds type hints plus matching docstring updates.  
I will submit the work as small, review-friendly PRs if approved.

PR #690, opened 2025-12-03
feat: Add type annotations to core API components

nishanthcr7777:
## What Does This PR Do?

This PR adds **standard Python type annotations** across core modules to improve code clarity, IDE support, and static analysis — without modifying any runtime logic.

---

## Files Updated

- `brainglobe_atlasapi/core.py`
- `brainglobe_atlasapi/utils.py`
- `brainglobe_atlasapi/structure_tree_util.py`
- `brainglobe_atlasapi/bg_atlas.py`
- `brainglobe_atlasapi/list_atlases.py`

---

### General Improvements
- Added **type hints** to all public-facing functions
- Corrected **misleading or outdated docstrings**
- **No runtime logic changed** (non-breaking)

---

## Why Is This Needed?

Although the project ships with `mypy` in dev dependencies, several core files were previously untyped. Adding type annotations:

- Improves **developer experience**
- Enables **better static analysis**
- Provides **stronger IDE autocompletion**
- Reduces **onboarding friction** for new contributors

---

#### Type Hints Added
- Return type annotations for:
  - `_version_tuple_from_str`
  - `_version_str_from_tuple`
- Fully typed the `BrainGlobeAtlas` class:
  - `__init__` constructor arguments
  - Properties:
    - `local_version`
    - `remote_version`
    - `local_full_name`
    - `remote_url`
  - Methods:
    - `download_extract_file`
    - `__repr__`
    - `__str__`

#### Type-safety Improvements
- Added `assert full_name is not None` in `local_version` to satisfy type checkers before string manipulation
- Refactored `remote_version` to use an intermediate variable for typed dictionary access

---

#### Type Hints Added
| Function | New Signature |
|----------|---------------|
| `get_downloaded_atlases` | `() -> List[str]` |
| `get_local_atlas_version` | `(atlas_name: str) -> Optional[str]` |
| `get_all_atlases_lastversions` | `() -> Dict[str, Any]` |
| `get_atlases_lastversions` | `() -> Dict[str, Dict[str, Any]]` |
| `show_atlases` | `(...) -> None` |
| `add_atlas_to_row` | `(...) -> Table` |

- Imported required types (`List`, `Dict`, `Any`, `Optional`) from `typing`

---

## Related Issue

Closes **#689**

---

## Testing

- `mypy` executed locally  
  - External stub-related warnings exist but are unrelated to this PR
- No change in runtime behavior

---

### DIFF
--- a/atlas_scripts/pre_packaging/drosophila/filtering_wingdisc.py
+++ b/atlas_scripts/pre_packaging/drosophila/filtering_wingdisc.py
@@ -6,10 +6,10 @@ improvement iterations.
 from pathlib import Path
 
 import numpy as np
-from brainglobe_template_builder.io import (
+from brainglobe_utils.io.image import (
+    load_nii,
     save_as_asr_nii,
 )
-from brainglobe_utils.IO.image import load_nii
 from skimage.filters.rank import modal
 from skimage.morphology import square
 
--- a/atlas_scripts/swc_female_rat.py
+++ b/atlas_scripts/swc_female_rat.py
@@ -1,0 +1,287 @@
+"""Package the swc_female_rat rat brain atlas (50 µm) in BrainGlobe format.
+
+This version:
+- Downloads template, annotation, and structures files from remote URLs
+- Uses annotation labels with Waxholm IDs
+"""
+
+__version__ = "0"
+
+import json
+from pathlib import Path
+
+import numpy as np
+import pooch
+import xmltodict
+from brainglobe_utils.IO.image import load_any
+
+from brainglobe_atlasapi import utils
+from brainglobe_atlasapi.atlas_generation.mesh_utils import (
+    construct_meshes_from_annotation,
+)
+from brainglobe_atlasapi.atlas_generation.wrapup import wrapup_atlas_from_data
+from brainglobe_atlasapi.config import DEFAULT_WORKDIR
+from brainglobe_atlasapi.structure_tree_util import get_structures_tree
+
+# -------------------------------------------------------------------------
+# Atlas metadata
+# -------------------------------------------------------------------------
+
+ATLAS_NAME = "swc_female_rat"
+SPECIES = "Rattus norvegicus"
+ATLAS_LINK = ""
+CITATION = "unpublished"
+ORIENTATION = "asr"
+RESOLUTION = (50, 50, 50)
+ROOT_ID = 10000  # Waxholm root
+ATLAS_PACKAGER = "Viktor Plattner, v.plattner@ucl.ac.uk"
+
+# File names on GIN.
+TEMPLATE_FILENAME = "template_sharpen_shapeupdate_orig-asr_aligned.nii.gz"
+ANNOTATION_FILENAME = "WHS_SD_annotation_template_space_cleaned.nii.gz"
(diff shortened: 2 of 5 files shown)


brainglobe_atlasapi/bg_atlas.py
```diff
@@ -3,7 +3,7 @@
 import tarfile
 from io import StringIO
 from pathlib import Path
-from typing import Optional
+from typing import Callable, Optional, Tuple, Union
```
IgorTatarnikov: Callable is [deprecated](https://docs.python.org/3/library/typing.html#typing.Callable)

Could you switch to using `collections.abc.Callable` https://docs.python.org/3/library/typing.html#annotating-callable-objects

brainglobe_atlasapi/bg_atlas.py
```diff
@@ -110,10 +110,11 @@ def local_version(self):
         if full_name is None:
             return None
 
+        assert full_name is not None
```
IgorTatarnikov: I don't think this assert is necessary, the `full_name is None` case is caught in the above if statement

brainglobe_atlasapi/bg_atlas.py
IgorTatarnikov: This function doesn't actually return `None`, an explicity `None` return should be added.

brainglobe_atlasapi/bg_atlas.py
```diff
@@ -154,7 +154,7 @@ def local_full_name(self):
             return candidate_dirs[0].name
 
     @property
-    def remote_url(self):
+    def remote_url(self) -> Optional[str]:
```
IgorTatarnikov, line 157: Same as above, an explicit None return should be added.

brainglobe_atlasapi/bg_atlas.py
IgorTatarnikov: We should add an explicit `return None` here.

brainglobe_atlasapi/structure_tree_util.py
IgorTatarnikov, line 83: This function is missing type hints.

The input is `Tree`, the output is `Generator[Node]`

brainglobe_atlasapi/list_atlases.py
```diff
@@ -12,7 +13,7 @@
 from brainglobe_atlasapi import config, descriptors, utils
 
 
-def get_downloaded_atlases():
+def get_downloaded_atlases() -> List[str]:
```
IgorTatarnikov, line 16: This should be a list of tuples of strings

brainglobe_atlasapi/list_atlases.py
```diff
@@ -12,25 +13,28 @@
...
 
     return [
-        f.name.rsplit("_v", 1)[0]
+        (
+            f.name.rsplit("_v", 1)[0],
+            re.search(r"_v(\d+\.\d+)$", f.name).group(1),
+        )
         for f in brainglobe_dir.glob("*_*_*_v*")
         if f.is_dir()
     ]
```
IgorTatarnikov: Sorry, I trused the docstring too much 😅! The previous implementation was correct and the docstring needs to be updated.

Could you revert the changes to the functionality, set the return as `List[str]` and udpate the docstring?

PR #700, opened 2025-12-07
Fix crash on incomplete atlas downloads (#698)

nishanthcr7777:
# PR Type

* [x] Bug fix
* [ ] Addition of a new feature
* [ ] Other

# Why is this PR needed?

Currently, if an atlas download is interrupted or results in a corrupted file, the application crashes with a `tarfile.ReadError` or `EOFError` during extraction because the downloaded file is not validated. This leaves a corrupted file on disk, preventing subsequent retries from succeeding without manual intervention. This PR addresses **Bug #698**.

# What does this PR do?

This PR introduces **integrity validation for downloaded atlas tarballs** before extraction:

* Wraps the extraction process in a `try-except` block to catch `tarfile.ReadError`, `CompressionError`, and `EOFError`.
* Automatically deletes corrupted files if validation fails, allowing clean retry downloads.
* Raises a descriptive, user-friendly error message when corruption is detected.
* Preserves the valid downloaded file upon successful extraction to ensure caching works as intended.

# References

* Fixes #698

# How has this PR been tested?

Two new test cases were added in `tests/atlasapi/test_bg_atlas.py`:

1. **`test_download_corrupted_file`**

   * Mocks a corrupted download (truncated gzip).
   * Verifies that `tarfile.ReadError` is raised.
   * Checks that the corrupted file is deleted from disk.

2. **`test_download_valid_file`**

   * Mocks a valid download.
   * Verifies extraction succeeds.
   * Confirms the original archive persists on disk.

All tests were run locally using `pytest`:

```bash
python -m pytest tests/atlasapi/test_bg_atlas.py -v
```

Result:  All tests passed

# Is this a breaking change?

No. It improves robustness of existing functionality.

# Does this PR require an update to the documentation?

No. This is an internal fix to handle errors more gracefully.

# Checklist

* [x] The code has been tested locally
* [x] Tests have been added to cover all new functionality
* [x] Documentation updated if necessary (N/A)
* [x] Code has been formatted using pre-commit

Issue #698, opened 2025-12-05
[BUG] demba_allen_seg_dev_mouse_p56 appears broken.


PolarBean:
**Describe the bug**
related to https://github.com/brainglobe/brainglobe-atlasapi/issues/697 each time I reinstall this atlas (twice now) It crashes during decompression. Strangely this occurs when the download is unfinished

<img width="1100" height="87" alt="Image" src="https://github.com/user-attachments/assets/6ced7409-b7c3-4f59-96b6-125018b62e2d" />

I am unsure whether this is specific to this atlas or my machine is just unstable which is causing the download to be interupted. Even so perhaps we should try to prevent the api from decompressing a file that is not completely downloaded. Perhaps by checking the pooch hash to ensure no corruption occured. 
**Computer used (please complete the following information):**

- OS: Windows - WSL

**Logs**

---------------------------------------------------------------------------
EOFError                                  Traceback (most recent call last)
Cell In[4], [line 7](vscode-notebook-cell:?execution_count=4&line=7)
      5 voxel_size_micron = 20
      6 space_name = r"demba_allen_seg_dev_mouse_p56"
----> [7](vscode-notebook-cell:?execution_count=4&line=7) atlas = BrainGlobeAtlas(f"{space_name}_{voxel_size_micron}um")
      8 source_age = 56
      9 target_age = 56

File ~/anaconda3/lib/python3.11/site-packages/brainglobe_atlasapi/bg_atlas.py:93, in BrainGlobeAtlas.__init__(self, atlas_name, brainglobe_dir, interm_download_dir, check_latest, config_dir, fn_update)
     91         raise ValueError(f"{atlas_name} is not a valid atlas name!")
     92     else:
---> [93](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/harryc/github/brainglobe-ccf-translator/~/anaconda3/lib/python3.11/site-packages/brainglobe_atlasapi/bg_atlas.py:93)         self.download_extract_file()
     95 # Instantiate after eventual download:
     96 super().__init__(self.brainglobe_dir / self.local_full_name)

File ~/anaconda3/lib/python3.11/site-packages/brainglobe_atlasapi/bg_atlas.py:181, in BrainGlobeAtlas.download_extract_file(self)
    179 # Uncompress in brainglobe path:
    180 tar = tarfile.open(destination_path)
[...]

PR #721, opened 2025-12-16
fix: recover from corrupted atlas metadata on initialization

### LINKED ISSUE #697: [BUG] crash when metadata.json is missing
**Describe the bug**
While installing an atlas python crashed. This left me with a mostly ok atlas folder which was missing metadata.json. 

This is of course recoverable if I manually delete the atlas folder and reinstall it. However this may be suboptimal when the atlasapi is embedded in other applications where the user might not understand this. 
**To Reproduce**
delete metadata.json from any atlas

**Expected behaviour**
Ideally it would autorecover by refetching the metadata.json file from the bg server. 

**Screenshots**

Image

**Computer used (please complete the following information):**

- OS: Windows 11 - WSL

nishanthcr7777:
### What is this PR
- [x] Bug fix

### Why is this PR needed?
If an atlas installation crashes during extraction, the local atlas directory can be left in a partially extracted state with a missing or corrupted `metadata.json`.  

Currently this causes `BrainGlobeAtlas` initialization to crash, requiring users to manually delete the atlas directory and reinstall ,which is not user-friendly.

### What does this PR do?
- Adds recovery logic during `BrainGlobeAtlas` initialization:
  - Detects missing or corrupted `metadata.json`
  - Removes the partially extracted atlas directory
  - Re-downloads the atlas once and retries initialization
- Uses a bounded retry loop (no recursion) to avoid infinite retries
- Leaves behavior unchanged for valid atlas installations

### References
Closes #697

### How has this PR been tested?
- Added a unit test that simulates a missing `metadata.json` by mocking `Atlas.__init__`
- Verified the atlas is re-downloaded and initialization succeeds
- Ran `pytest` locally and confirmed existing tests pass

### Is this a breaking change?
No.  
This change only affects recovery from corrupted atlas installs.

### Does this PR require an update to the documentation?
No.  

## Checklist:
- [x] Tests have been added to cover the new recovery behavior
- [x] The code has been formatted with pre-commit

### DIFF
--- a/brainglobe_atlasapi/bg_atlas.py
+++ b/brainglobe_atlasapi/bg_atlas.py
@@ -1,12 +1,13 @@
 """Defines the BrainGlobeAtlas class for accessing brain atlas data."""
 
+import shutil
 import tarfile
+import warnings
 from collections.abc import Callable
 from io import StringIO
 from pathlib import Path
 from typing import Optional, Tuple, Union
-import shutil
-import warnings
+
 import requests
 from rich import print as rprint
 from rich.console import Console
@@ -118,7 +119,7 @@ class BrainGlobeAtlas(core.Atlas):
 
         if check_latest:
             self.check_latest_version()
-    
+
     @property
     def local_version(self) -> Optional[Tuple[int, ...]]:
         """If atlas is local, return actual version of the downloaded files;


alessandrofelder (commented):
Thanks @nishanthcr7777 - could you help me understand how this PR relates to your other PR #700, please?

It looks to me like both are meant to address #697 and #698 (but #698 is a special case of #697). Do they solve the same problem or just a very similar problem?

Do we need both PRs? And in which order?

brainglobe_atlasapi/bg_atlas.py
```diff
@@ -96,7 +98,24 @@ def __init__(
                 self.download_extract_file()
 
         # Instantiate after eventual download:
-        super().__init__(self.brainglobe_dir / self.local_full_name)
+        for attempt in (0, 1):
```
PolarBean, line 101: why not `range(2)`?

n1shanthb: Mostly for readability, I wanted to make it explicit that this is exactly two attempts (initial + one retry). I’m happy to switch to range(2) if that’s preferred.

PolarBean: all good

Issue #233, opened 2023-04-17
Ensure Allen mouse brain atlas is up to date


adamltyson:
Based on [this post](https://community.brain-map.org/t/2022-ccfv3-mouse-atlas/2287) it seems that there are updated versions of the annotations. Should we be releasing a new version of the atlas (and downsampling ourselves)?

PolarBean:
Yeah for the 2022 version it seems there are still no provided downsamples.

PolarBean:
I replicated the allen downsampling for the 2017 version, this was previously an issue as it wasnt clear how they were downsampling their volumes. If we apply this to the 2022 10um we can make new dowsampled atlases which maintain consistent region volumes in the downsamples for non updated regions. 

```python
import nrrd
import requests
import numpy as np

def download_to_file(url, filename):
    response = requests.get(url)
    with open(filename, "wb") as file:
        file.write(response.content)

def float_downsample_alternating(volume, pattern):
    def make_indices(max_dim, pattern):
        idx, current = [], 0
        i = 0
        while current < max_dim:
            idx.append(current)
            current += pattern[i % len(pattern)]  # cycle over entire pattern
            i += 1
        return np.array(idx)

    idx_z = make_indices(volume.shape[0], pattern)
    idx_y = make_indices(volume.shape[1], pattern)
    idx_x = make_indices(volume.shape[2], pattern)

    return volume[np.ix_(idx_z, idx_y, idx_x)]


url_100um_allen = r"https://download.alleninstitute.org/informatics-archive/current-release/mouse_ccf/annotation/ccf_2017/annotation_100.nrrd"
url_50um_allen = r"https://download.alleninstitute.org/informatics-archive/current-release/mouse_ccf/annotation/ccf_2017/annotation_50.nrrd"
url_25um_allen = r"https://download.alleninstitute.org/informatics-archive/current-release/mouse_ccf/annotation/ccf_2017/annotation_25.nrrd"
url_10um_allen = r"https://download.alleninstitute.org/informatics-archive/current-release/mouse_ccf/annotation/ccf_2017/annotation_10.nrrd"
download_to_file(url_100um_allen, "annotation_100.nrrd")
download_to_file(url_50um_allen, "annotation_50.nrrd")
download_to_file(url_25um_allen, "annotation_25.nrrd")
download_to_file(url_10um_allen, "annotation_10.nrrd")
allen_100, header_100 = nrrd.read("annotation_100.nrrd")
[...]

PR #727, opened 2025-12-17
update script for 2022 version of allen, remove allensdk

PolarBean:
## Description
Here we add the 2022 version of the Allen. We also remove the allensdk as it is becoming increasingly incompatible with other parts of brainglobe (getting it to work in Python3.13 while supporting threading was too hard)
**What is this PR**

- [ ] Bug fix
- [ ] Other

## References
#233 

Yes we will update the atlas docs. 
## Checklist:


PolarBean:
25 micron file
https://data-proxy.ebrains.eu/api/v1/buckets/deepslice/allen_mouse_25um_v1.3.tar.gz

PolarBean:
The allen have done a really fantastic job making smooth meshes. I cant really get comparable quality. I'll just reuse their meshes and restrict the mesh generation to regions that have been updated. Does anyone know if they have open sourced their mesh generation pipeline as I think it could be really useful, for other atlases also.

PolarBean:
I made some change to the mesh generation script but its quite rigid in that it expects a single tuple arg parameter (i assume this is for multithreading purposes). Can we update the function create_region_mesh to accept key word arguments instead of a tuple. There are ways to make this work with multithreading. Since this function is used in some existing atlas packaging scripts, for backwards compatiblility purposes we could for the moment have a new_create_region_mesh and create_region_mesh would take a tuple and feed it to the new function. 

We could then update previous atlas scripts and the tests to the new function (which would be straight forward). Or we could update those scripts in this PR and get rid of the tuple version of the function. Just let me know

adamltyson:
Hey @PolarBean, probably best to create a new issue and we can discuss. The mesh generation functions are long overdue some more thought. @brainglobe is on holiday for two weeks now though, so we'll take a look at this PR in Jan.

PolarBean:
This is now split into two PRs with the mesh refactoring now at #761. I have also regenerated everything at all four resolutions (10, 25, 50, and 100). I'll upload them to gin once this passes review.

atlas_scripts/allen_mouse.py
```diff
@@ -1,140 +1,309 @@
...
+import csv
+import json
 
-from allensdk.api.queries.ontologies_api import OntologiesApi
-from allensdk.api.queries.reference_space_api import ReferenceSpaceApi
-from allensdk.core.reference_space_cache import ReferenceSpaceCache
-from requests import exceptions
-from tqdm import tqdm
+import meshio as mio
+import nrrd
```
IgorTatarnikov, line 14: This should be added to the `atlasgen` dependency group

IgorTatarnikov (commented):
The meshes for the 10um atlas aren't scaled correctly, see screenshot below:

Image

This doesn't seem to be the case for the 25um atlas.

atlas_scripts/allen_mouse.py
PolarBean: is it better to add it to the atlasgen dependency or readd the allenmouse dependency group?

PolarBean:
Is this just waiting for the wrapup v2 or should I change anything?

PolarBean:
This passes for me. I generated the api v3 version which has no issues other than the mesh coordinate violation which we previous identified was from a random unmeshed pixel and not a scale problem. 

For v3 should I regenerate at multiple resolutions or just the highest?

PR #732, opened 2025-12-23
add CArea atlas generation script

PolarBean:
This adds the unpublished Clustered Area (CArea) atlas. There will soon be a preprint and I will update the script with that DOI once this is out. 

I have validated this works in napari and that the template is within the mesh extent. 
image

One thing is it fails a validation function saying that a pixel is more than 10 times the mesh extent. I'm not sure what this is about but it could be a quirk of this atlas, given that the regions are noncontiguous. Everything looks good in napari.
image

### DIFF
--- a/atlas_scripts/carea_mouse.py
+++ b/atlas_scripts/carea_mouse.py
@@ -27,7 +27,7 @@ from brainglobe_atlasapi.atlas_generation.wrapup import wrapup_atlas_from_data
 # if this is the first time this atlas has been added the value should be 0
 # (minor version is the first number after the decimal point, ie the minor
 # version of 1.2 is 2)
-__version__ = 0
+__version__ = 1
 
 # The expected format is FirstAuthor_SpeciesCommonName, e.g. kleven_rat, or
 # Institution_SpeciesCommonName, e.g. allen_mouse.
@@ -36,7 +36,7 @@ ATLAS_NAME = "carea_mouse"
 # DOI of the most relevant citable document
 # DOI or URL of the most relevant citable document.
 # If no paper/DOI exists, use "unpublished".
-CITATION = "unpublished"
+CITATION = "https://doi.org/10.64898/2026.01.20.700446"
 
 # The scientific name of the species, ie; Rattus norvegicus
 SPECIES = "Mus musculus"


PolarBean:
https://data-proxy.ebrains.eu/api/v1/buckets/deepslice/carea_mouse_25um_v1.0.tar.gz

See the atlas files here.

adamltyson:
Hey @PolarBean, a couple of things:
- If this isn't published, should merging this wait until after publication, or is the idea that it's released on BG as part of the publication?
- @IgorTatarnikov seems to think the mesh issue is because the meshes are generated in pixel space, not micron space. In napari you don't have any meshes loaded (double click on the region in the bottom right panel), so they may not overlap properly.

atlas_scripts/carea_mouse.py
```diff
@@ -0,0 +1,331 @@
...
+__version__ = 0
+
+# The expected format is FirstAuthor_SpeciesCommonName, e.g. kleven_rat, or
+# Institution_SpeciesCommonName, e.g. allen_mouse.
+ATLAS_NAME = "carea_mouse"
+
+# DOI of the most relevant citable document
+# DOI or URL of the most relevant citable document.
+# If no paper/DOI exists, use "unpublished".
+CITATION = "unpublished"
```
adamltyson: isn't this published now?

atlas_scripts/carea_mouse.py
```diff
@@ -0,0 +1,331 @@
...
+# Copy-paste this script into a new file and fill in the functions to package
+# your own atlas.
+
+### Metadata ###
+
+# The minor version of the atlas in the brainglobe_atlasapi, this is internal,
+# if this is the first time this atlas has been added the value should be 0
+# (minor version is the first number after the decimal point, ie the minor
+# version of 1.2 is 2)
+__version__ = 0
```
adamltyson: needs to be bumped once the citation is added

atlas_scripts/carea_mouse.py
PolarBean: yes!, added the doi now

atlas_scripts/carea_mouse.py
PolarBean: bumped

PR #733, opened 2026-01-01
Add early atlas output check with overwrite option

### LINKED ISSUE #731: [Feature] catch existing atlas files early in atlas generation script rather than at the end
Embarrasingly often I get the following error after running a long atlas generation script. 
I understand the desire to not automatically overwrite files but we could potentially move this to the beginning of the atlas generation template rather than occuring at the end. 

```
  File "/home/harryc/github/brainglobe-atlasapi/brainglobe_atlasapi/bg_atlas.py", line 88, in __init__
    if self.local_full_name is None:
       ^^^^^^^^^^^^^^^^^^^^
  File "/home/harryc/github/brainglobe-atlasapi/brainglobe_atlasapi/bg_atlas.py", line 145, in local_full_name
    raise FileExistsError(
FileExistsError: Multiple versions of atlas carea_mouse_25um in /home/harryc/brainglobe_workingdir/carea_mouse
```

We could also add an argument here which allows the atlas generator to overwrite existing files https://github.com/brainglobe/brainglobe-atlasapi/blob/main/brainglobe_atlasapi/atlas_generation/wrapup.py#L288-L292

AlgoFoe:
## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**
- Atlas generation could fail only at the very end if an atlas output directory already existed, this PR ensures such cases are detected early.

**What does this PR do?**
- Adds an early check for existing atlas output directories in ```wrapup_atlas_from_data``` and introduces an overwrite option to explicitly replace them when needed.

## References
- Fixes #731 

- Tested locally by simulating existing atlas output directories.

## Is this a breaking change?
- No, default behavior remains unchanged and the new overwrite flag is optional

- No

## Checklist:

### DIFF
--- a/tests/atlasgen/test_wrapup_overwrite.py
+++ b/tests/atlasgen/test_wrapup_overwrite.py
@@ -12,6 +12,7 @@ from brainglobe_atlasapi.atlas_generation.wrapup import (
     wrapup_atlas_from_data,
 )
 
+
 def _minimal_valid_inputs(tmp_path):
     """
     Return a minimal set of valid inputs required to run wrapup


adamltyson:
Hey @AlgoFoe, thanks for this. While this is useful, it doesn't entirely fix the issue. Most of the time consuming steps take place before this wrap up script is run. 

@PolarBean's original issue mentions adding a check to the start of the [template script](https://github.com/brainglobe/brainglobe-atlasapi/blob/main/atlas_scripts/template_script.py) so the issue can be caught earlier. Would you mind adding something there too?

AlgoFoe:
Just pushed the requested updates and also added minimal tests to cover the ```wrapup``` output-check logic.

adamltyson (approved):
Great, once again, thank you @AlgoFoe!

PR #734, opened 2026-01-03
feat: add validation for duplicate structure acronyms

### LINKED ISSUE #666: [Feature] Check for duplicate acronyms during validation of atlases
Duplicate acronyms for region names are incompatible with the current implementation of `brainglobe-atlasapi` as we use the acronym as a primary key to fetch details for the region.

The presence of duplicate annotations should throw an error during atlas wrapup.

thisisrick25:
## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**

Closes #666

Duplicate structure acronyms are incompatible with the current implementation of ``brainglobe-atlasapi`` as acronyms are used as primary keys to fetch region details. This validation catches such issues during atlas generation.

**What does this PR do?**

Adds a [validate_unique_acronyms](cci:1://file:///c:/Users/swapn/code/brainglobe/brainglobe-atlasapi/brainglobe_atlasapi/atlas_generation/validate_atlases.py:426:0-458:15) function that checks for duplicate acronyms in atlas structures during validation and raises an `AssertionError` with a meaningful error message listing the duplicate acronyms.

## References

- Issue #666

- Added ``test_validate_unique_acronyms_fail`` test that mocks atlas structures with duplicate acronyms and verifies the validation fails with the expected error message
- The existing ``test_valid_atlas_passes_all_validations`` test will automatically verify the function passes on valid atlases

## Is this a breaking change?

No!

No, this is an internal validation function used during atlas generation.

## Checklist:

### DIFF
--- a/brainglobe_atlasapi/atlas_generation/validate_atlases.py
+++ b/brainglobe_atlasapi/atlas_generation/validate_atlases.py
@@ -459,7 +459,7 @@ def validate_unique_acronyms(atlas: BrainGlobeAtlas):
 
     assert (
         len(duplicates) == 0
-    ), f"Duplicate acronyms found in atlas structures: {duplicates}"
+    ), f"Duplicate acronyms found in atlas structures: {sorted(duplicates)}"
     return True
 
 


brainglobe_atlasapi/atlas_generation/validate_atlases.py
```diff
@@ -424,6 +424,41 @@ def validate_annotation_symmetry(atlas: BrainGlobeAtlas):
...
+    Raises
+    ------
+    AssertionError
+        If any duplicate acronyms are found in the atlas structures.
+    """
+    acronyms = [
+        atlas.structures[structure]["acronym"]
+        for structure in atlas.structures
+    ]
+    duplicates = [acr for acr in acronyms if acronyms.count(acr) > 1]
```
IgorTatarnikov: This can be rewritten slightly to improve performance. We don't need the specific count for each acronym, we only care if it's a duplicate.

If we build the set one by one, we can check if each acronym already exists in the set (this is a constant time operation, and doesn't scale with the set length). This way we loop over the list of acronyms once, construct the set, and have a list of possibly duplicated values to report.

brainglobe_atlasapi/atlas_generation/validate_atlases.py
```diff
@@ -424,6 +424,41 @@ def validate_annotation_symmetry(atlas: BrainGlobeAtlas):
...
+    unique_duplicates = list(set(duplicates))
```
IgorTatarnikov: It might be useful to report a tuple where the first element is the acronym, and the second element is the full name. For any acronyms that are duplicates you can access the the full name using `atlas.structures[acronym]['name']`

PR #742, opened 2026-02-02
repackage after update doi of allen cord

PolarBean:
after updating the doi we needed to repackage the spinal cord. Heres is the regenerated file. 

Also! the previous script include a direct link to the data on aws not via mendeley. It seems the aws data is temporary and expires. Better to use the mendeley link.


PolarBean:
I have tested in napari and uploaded the files to gin!

PR #745, opened 2026-02-03
add atlas name hinting

PolarBean:
## Description
I always find myself needing to check the brainglobe documentation for atlas names. I realised it would be really nice if my IDE could suggest me valid atlas names. This is apparantly very doable and I have implemented it here. The main drawback is we would have a hardcoded atlas name list that we would need to update. We could have a test though that automatically checks for discrepencies between the remote last_versions.conf and our atlas_names.py

This is what it looks like in practice
image

**What is this PR**

- [ ] Bug fix
- [ ] Other

## Checklist:

### DIFF
--- a/atlas_scripts/template_script.py
+++ b/atlas_scripts/template_script.py
@@ -22,7 +22,8 @@ __version__ = 0
 
 # The expected format is FirstAuthor_SpeciesCommonName, e.g. kleven_rat, or
 # Institution_SpeciesCommonName, e.g. allen_mouse.
-# remember to add {ATLAS_NAME}_{RESOLUTION}um to brainglobe_atlasapi/atlas_names.py
+# remember to add {ATLAS_NAME}_{RESOLUTION}um to:
+# brainglobe_atlasapi/atlas_names.py
 ATLAS_NAME = "example_mouse"
 
 # DOI of the most relevant citable document


PolarBean:
I think its odd to change the python files after everything is installed. Here is what the gh action solution would look like, what do you think?

tests/atlasapi/test_list_atlases.py
```diff
@@ -79,6 +81,21 @@ def test_get_all_atlases_lastversions():
     assert "allen_mouse_25um" in last_versions
 
 
+def test_atlas_name_matches_lastversions():
+    """Ensure atlas name list matches last_versions.conf keys exactly."""
+    atlas_name_values = list(get_args(AtlasName))
+    cache_path = config.get_brainglobe_dir() / "last_versions.conf"
+    if not cache_path.exists():
```
alessandrofelder: I think we should make sure that our test env always has this set up, so would shouldn't need to check this?
I think we (kind of) already implicitly do because we have [a fixture that instantiates a BrainGlobeAtlas](https://github.com/brainglobe/brainglobe-atlasapi/blob/14affb7bf470103ccec5c00b7dc8b906a4f962bc/tests/conftest.py#L14) that will have been called previously.

alessandrofelder:
I like the idea a lot!
Maybe I am too tired :sweat_smile: will this work alongside users that have [a custom_atlases.conf](https://github.com/brainglobe/brainglobe-atlasapi/blob/14affb7bf470103ccec5c00b7dc8b906a4f962bc/brainglobe_atlasapi/list_atlases.py#L62) file?

IIUC they just wouldn't have the IDE autocomplete but they could still instantiate the atlas objects?

tests/atlasapi/test_list_atlases.py
PolarBean: I agree its a bit extra, ill remove it

PR #751, opened 2026-02-12
pin vtk to 9.6.0 before GetCells is deprecated

PolarBean:
tests have begun failing. I think it is due to a dependancy being updated and us relying on deprecated functions.


IgorTatarnikov:
Looks like this is fixed as of https://github.com/marcomusy/vedo/releases/tag/v2026.6.1, see https://github.com/brainglobe/brainglobe-atlasapi/actions/runs/21959269947.

I think this can be closed now.

Issue #764, opened 2026-03-02
[BUG] Invalid RGB triplet for region 18 (Cgus) in azba_zfish_4um


IgorTatarnikov:
**Describe the bug**
The RGB triplet for the area 18 (Cgus) is invalid in azba_zfish_4um. The `structure.json` entry for this region is `{"acronym": "Cgus", "id": 18, "name": "commissure of the secondary gustatory nuclei", "structure_id_path": [9999, 18], "rgb_triplet": [15, 2, 2003]}`. 2003 is an invalid value.

I traced this back to the original data source. The BrainGlobe values match the values in `2021-08-22_AZBA_labels.csv`.

**Expected behaviour**
The `rgb_triplet` entry should be valid.

HARSHDIPSAHA:
> I traced this back to the original data source. The BrainGlobe values match the values in `2021-08-22_AZBA_labels.csv`.

yes i verified this,by running,

``python3 -c "
import csv, tarfile, urllib.request, io
url = 'http://www.azba.wayne.edu/2021-08-22_AZBA.tar.gz'
print('Fetching source CSV...')
data = urllib.request.urlopen(url).read()
with tarfile.open(fileobj=io.BytesIO(data), mode='r:gz') as tf:
    m = tf.extractfile('2021-08-22_AZBA_labels.csv')
    rows = list(csv.DictReader(io.TextIOWrapper(m, encoding='utf-8')))
row18 = next(r for r in rows if r['id'] == '18')
print(f'Structure: {row18['name']}')
print(f'Raw CSV Value: {row18['rgb_triplet']}')
``

i got Raw CSV Value: 15/2/2003

Issue #763, opened 2026-03-02
[BUG] Mesh for 614454502 is invalid for `allen_mouse_bluebrain_10um`


IgorTatarnikov:
**Describe the bug**
The file `614454502.obj` is invalid for `allen_mouse_bluebrain_10um`. The values are far too large.

```
# Created by meshio v5.3.4, 2024-08-02T09:45:05.638719
v 7.019848e+234 8.591699999999986e+233 2.557681100000002e+234
v 7.02987119999999e+234 8.591699999999986e+233 2.557681100000002e+234
v 7.026799299999999e+234 8.55610349999999e+233 2.564992999999999e+234
v 7.019563000000001e+234 8.559064500000002e+233 2.5672314000000062e+234
v 7.032811899999998e+234 8.645332299999999e+233 2.5636440999999995e+234
v 7.009824800000001e+234 8.591699999999986e+233 2.5677057e+234
v 7.0115771e+234 8.548443600000007e+233 2.5743414e+234
v 7.024407299999999e+234 8.5216446e+233 2.573182400000004e+234
v 7.039895000000007e+234 8.591699999999986e+233 2.5677057e+234
v 7.034726000000001e+234 8.532846100000007e+233 2.5704279000000012e+234
v 7.051111500000002e+234 8.496111300000001e+233 2.5783640000000043e+234
v 7.064626500000002e+234 8.611148099999998e+233 2.572664800000001e+234
v 7.024135700000003e+234 8.685633900000001e+233 2.5663271999999983e+234
v 7.012036700000001e+234 8.717110399999988e+233 2.5703447999999997e+234
v 7.077072799999995e+234 8.65842359999999e+233 2.574857200000001e+234
v 7.078328199999998e+234 8.585939000000003e+233 2.574586800000002e+234
v 7.066597299999998e+234 8.678710199999994e+233 2.5713918999999995e+234
```

**To Reproduce**
Try loading the mesh for 614545502.

**Expected behaviour**
The mesh file should be valid.

PR #769, opened 2026-03-11
BrainGlobeAtlas API V2 - core

### LINKED ISSUE #245: Consider compressing the meshes
`.obj` works well, but the `mesh` directories can get large. Using newer formats could improve this a lot. 

[Draco](https://google.github.io/draco/) looks promising.

### LINKED ISSUE #291: Save atlases to `.brainglobe/atlases`
We now have `.brainglobe/cellfinder` and `.brainglobe/brainrender`. We should create `.brainglobe/atlases` so that directory is easier to navigate. 

Part of https://github.com/brainglobe/BrainGlobe/issues/26 and linked to https://github.com/brainglobe/brainglobe-atlasapi/issues/284

### LINKED ISSUE #407: Speed up atlas downloads
Downloading the atlases can take a long time. In some areas it can seemingly take hours. We should investigate alternatives to improve this, e.g.:
- Move to other providers (e.g. Zenodo https://github.com/brainglobe/brainglobe-atlasapi/issues/284#issuecomment-2335129409)
- Region-specific mirrors (cc @dbirman)

IgorTatarnikov:

## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**
Atlases are currently fetched as monolithic tar archives that must be downloaded in entirety before the atlas can be used.

This PR separates the atlas into the corresponding components and only fetches the components as they are requested. Installing an atlas just downloads all of the relevant metadata. This simplifies the installation process and allows the reuse of components by multiple atlases.

This PR also moves the primary download directory from GIN to S3. This should increase download speeds and reliability for downstream users.

https://atlas-assets.readthedocs.io/en/latest/ was used as the guide for structuring the new atlases.

**What does this PR do?**
- Introduces BrainGlobeAtlas API V2 that fetches atlases from the BrainGlobe S3 bucket in OME-Zarr / NGFF format, replacing the GIN-hosted .tar.gz archives used by the legacy API.
- Extracts legacy behaviour into a new `BrainGlobeAtlasLegacy` class (bg_atlas_legacy.py), keeping backward-compatibility intact for users who wish to continue using legacy atlases.
- Implements transparent fallback: a `_FallbackToLegacyMeta` metaclass on `BrainGlobeAtlas` silently retries with `BrainGlobeAtlasLegacy` when an atlas raises `AtlasNotAvailableAsV2`, and ensures `isinstance(obj, BrainGlobeAtlas)` returns `True` for legacy instances. Callers can also opt-in explicitly with `legacy=True`.
- Updates `core.Atlas` to support both v1 (local directory) and v2 (S3 path) atlas layouts, including separate pyramid levels for template and annotation volumes, hemisphere initialisation on download, and additional reference fetching.
- Switches atlas discovery to S3: `list_atlases` and `get_all_atlases_lastversions` now read last_versions.conf from the S3 bucket (via HTTP), falling back to a local cache when offline.
- Updates all tests to match the new structure; validation tests are marked xfail pending finalisation of the v2 spec.
- Version identifiers use _ as separator in v2 (e.g. 1_0) vs . in v1 (e.g. 1.0).

Note:

V2 atlases use OME-Zarr multiscale arrays on S3 (ngff_zarr + s3fs); pyramid level is resolved from the atlas resolution metadata. This is temporary solution that eases the transition, allowing users to continue using the current BrainGlobe atlas names (e.g. `allen_mouse_25um`). In the future, we will remove the resolution suffix from the atlas name and add an optional `resolution` arguments to the `__init__` of `BrainGlobeAtlas`. Without the resolution parameter the user could either receive the atlas at the highest available resolution, or receive all resolutions (however this would be a breaking change as we will no longer guarantee that `atlas.annotations` is a 3D array).

To do:
- Update the packaging and validation functions to account for BrainGlobeAtlas API V2
- Add support for an expanded 4D annotation array that stores the mask for each label
- Add more metadata to existing atlases and expose it via the API
- Add validators for the metadata of each component to ensure a valid atlas collection

## References
#141 #562

Tested locally by visual and programmatic means.

I used the below script to compare two directories, one with existing legacy atlases, and a new directory containing the newly packaged counterparts. Both directories can point to the current BrainGlobe dir to test against currently installed legacy atlases. Any failures from this comparison were either related to issues in the existing legacy atlas #764 #763, or the scaling of the meshes. All meshes are now stored at the highest resolution, while some of the existing legacy atlases package meshes at multiple resolutions.

<details>
<summary>Comparison Script</summary>

```
from concurrent.futures import ProcessPoolExecutor
from pathlib import Path

import numpy as np
import meshio

from brainglobe_utils.general.system import get_num_processes
from brainglobe_atlasapi import BrainGlobeAtlas

brainglobe_legacy_dir = Path("/mnt/Data/.brainglobe")
brainglobe_dir = Path("/mnt/Data/.brainglobe_v2_final")

atlas_glob = brainglobe_legacy_dir.glob('*_*um_v*')
legacy_atlas_paths = [path.name for path in atlas_glob]
failed_atlases = []

def compare_structures(structure1, structure2):
    assert structure1["name"] == structure2["name"], f"Structure names do not match: {structure1['name']} vs {structure2['name']}"
    assert structure1["acronym"] == structure2["acronym"], f"Structure acronyms do not match: {structure1['acronym']} vs {structure2['acronym']}"
    assert structure1["rgb_triplet"] == structure2["rgb_triplet"], f"Structure colors do not match: {structure1['rgb_triplet']} vs {structure2['rgb_triplet']}"
    assert structure1["structure_id_path"] == structure2["structure_id_path"], f"Structure paths do not match: {structure1['structure_id_path']} vs {structure2['structure_id_path']}"
    try:    
        old_mesh = structure1["mesh"]
        new_mesh = structure2["mesh"]
        if old_mesh is not None and new_mesh is not None:
            assert np.allclose(old_mesh.points, new_mesh.points), f"Mesh points do not match for structure {structure1['id']}"
    except meshio.ReadError:
        pass # If the mesh cannot be read, we skip the comparison for this structure

for atlas_dir in legacy_atlas_paths:
    atlas_name = "_".join(atlas_dir.split("_")[:-1])

    print(f"Comparing {atlas_name}...")
    try:
        new_atlas = BrainGlobeAtlas(atlas_name, brainglobe_dir=brainglobe_dir, check_latest=False)
        legacy_atlas = BrainGlobeAtlas(atlas_name, brainglobe_dir=brainglobe_legacy_dir, legacy=True,
                                       check_latest=False)
        assert legacy_atlas.atlas_name == new_atlas.atlas_name, f"Atlas names do not match: {legacy_atlas.atlas_name} vs {new_atlas.atlas_name}"
        assert legacy_atlas.shape == new_atlas.shape, f"Atlas shapes do not match: {legacy_atlas.shape} vs {new_atlas.shape}"
        assert legacy_atlas.resolution == new_atlas.resolution, f"Atlas resolutions do not match: {legacy_atlas.resolution} vs {new_atlas.resolution}"
        assert len(legacy_atlas.structures) <= len(
            new_atlas.structures), f"Number of structures do not match: {len(legacy_atlas.structures)} vs {len(new_atlas.structures)}"
        assert np.all(
            legacy_atlas.annotation == new_atlas.annotation), f"Atlas annotations do not match for {atlas_name}"
        assert np.all(legacy_atlas.reference == new_atlas.template), f"Atlas templates do not match for {atlas_name}"
        assert np.all(
            legacy_atlas.hemispheres == new_atlas.hemispheres), f"Atlas hemispheres do not match for {atlas_name}"
        num_processes = get_num_processes()

        # Use a process pool to compare meshes in parallel
        with ProcessPoolExecutor(max_workers=num_processes) as executor:
            futures = [executor.submit(compare_structures, legacy_atlas.structures.get(s["id"]), new_atlas.structures.get(s["id"]))
                       for s in legacy_atlas.structures_list]
            
            for future in futures:
                future.result()  # This will raise an exception if the comparison failed

        print(f"✅ {atlas_name} passed.")
    except AssertionError as e:
        print(f"❌ {atlas_name} failed: {e}")
        failed_atlases.append((atlas_name, str(e)))
    except Exception as e:
        print(f"❌ {atlas_name} failed with unexpected error: {e}")
        failed_atlases.append((atlas_name, str(e)))

    print("" + "-" * 50 + "\n")

print("Comparison complete.")
if failed_atlases:
    print("The following atlases failed the comparison:")
    for atlas in failed_atlases:
        print(f"- {atlas}")
else:
    print("All atlases passed the comparison.")
```

</details>

## Is this a breaking change?
No.

Yes, this should be followed up with a blog post and updates to https://brainglobe.info 

## Checklist:


brainglobe_atlasapi/update_atlases.py
```diff
@@ -39,22 +39,25 @@ def update_atlas(atlas_name, force=False, fn_update=None):
                 "[/b]"
             )
             return
+    else:
+        if atlas.check_latest_version(print_warning=False):
```
PolarBean: if force is False fld is never defined. line 53 fld.exists will raise an error. 

Also i find the dobule negative of "if not force" a bit confusing, why not just switch the code blocks and have if force else

PolarBean: this line really confuses me as it seems like check_latest_version returns true if it is the latest version of the atlas. In that case this line only deletes the atlas folder if it is the latest version and does nothing if the atlas is not updated. 

This seems problematic for two reasons, one this is the inverse of what I would expect. second i would expect force to always delete the atlas

brainglobe_atlasapi/list_atlases.py
```diff
@@ -23,12 +22,17 @@ def get_downloaded_atlases() -> List[str]:
...
+    atlases_dir = brainglobe_dir / "atlases"
 
-    return [
-        f.name.rsplit("_v", 1)[0]
-        for f in brainglobe_dir.glob("*_*_*_v*")
-        if f.is_dir()
-    ]
+    downloaded_atlases = []
+
+    for f in atlases_dir.iterdir():
```
PolarBean, line 34: when is the atlases dir created? if someone runs this function on a fresh install before any atlases exist will this crash?

brainglobe_atlasapi/structure_class.py
```diff
@@ -44,22 +47,44 @@ def __getitem__(self, item):
...
                 )
 
         return self.data[item]
 
+    def _check_mesh_cached(self, file_name: Path):
+        """Check if the mesh is cached, and if not, attempt to load it."""
+        if file_name.exists():
+            return
+
+        root_path = "/".join(str(file_name).split("/")[-5:])
```
PolarBean: i dont think you should hard code the seperator like this as i think this will crash on windows. Try os.sep instead

brainglobe_atlasapi/core.py
```diff
@@ -119,20 +206,71 @@ def lookup_df(self):
...
-        Loads it if not already loaded.
+    def reference(self):
+        """Return the template image data.
+
+        Warning: this is a deprecated alias for template, and will be removed
+        in future versions. Use atlas.template instead.
         """
-        if self._annotation is None:
-            self._annotation = read_tiff(self.root_dir / ANNOTATION_FILENAME)
+        print(
```
PolarBean, line 220: i think this should be warnings.warn

brainglobe_atlasapi/core.py
```diff
@@ -119,20 +206,71 @@ def lookup_df(self):
...
+        annotation_location = self.metadata["annotation_set"]["location"][1:]
+        annotation_path = (
+            self.root_dir / annotation_location / V2_ANNOTATION_NAME
+        )
+
+        multiscale = nz.from_ngff_zarr(annotation_path)
+        resolution_path = annotation_path / str(self._annotation_pyramid_level)
+
+        if not (resolution_path / "c").exists():
+            print("Downloading annotations...")
```
PolarBean, line 241: why not use the cool progress bars we used to have

brainglobe_atlasapi/descriptors.py
```diff
@@ -47,6 +49,19 @@
 HEMISPHERES_FILENAME = "hemispheres.tiff"
 MESHES_DIRNAME = "meshes"
 
+# V2 file names
+V2_ATLAS_ROOTDIR = "atlases"
+V2_ANNOTATION_ROOTDIR = "annotation-sets"
+V2_COORDINATE_SPACE_ROOTDIR = "coordinate-spaces"
+V2_TEMPLATE_ROOTDIR = "templates"
+V2_TERMINOLOGY_ROOTDIR = "terminologies"
+V2_TERMINOLOGY_NAME = "terminology.csv"
```
PolarBean, line 58: this is idential to V2_STRUCTURES_NAME, perhaps just use one

tests/atlasgen/test_validation.py
```diff
@@ -166,6 +166,10 @@ def atlas_with_reference_matching_additional_reference():
     os.remove(additional_reference_name)
 
 
+@pytest.mark.xfail(
+    reason="This test is currently failing as the validation functions "
```
PolarBean, line 170: is there an issue to track these tests so we remember to come back to them

brainglobe_atlasapi/bg_atlas.py
```diff
@@ -75,145 +128,269 @@ def __init__(
...
+            f"{V2_ATLAS_ROOTDIR}/{self.atlas_name}/"
+            f"{remote_version_str}/manifest.json"
         )
 
-        # Uncompress in brainglobe path:
+        local_path = self.brainglobe_dir / key_name
+        remote_path = remote_url_s3.format(key_name)
+
+        local_path.parent.mkdir(parents=True, exist_ok=True)
+        self.fs.get(remote_path, local_path)
```
PolarBean: is this writing the manifest file? if so it contradicts the docstring that the manifest writing happens last

brainglobe_atlasapi/update_atlases.py
```diff
@@ -39,22 +39,25 @@ def update_atlas(atlas_name, force=False, fn_update=None):
...
+
     if fld.exists():
         raise ValueError(
             "Something went wrong while trying to delete the old version "
             "of the atlas, aborting."
         )
 
     # Download again
-    atlas.download_extract_file()
+    atlas.download()
```
PolarBean, line 60: after reading atlas.download it doesnt seem like it updates local_full_name? so if i update the atlas will local_full_name still point to the old deleted directory?

brainglobe_atlasapi/update_atlases.py
IgorTatarnikov: Good catch, I'm going to remove the `fld` exists check and flip the conditionals.

IgorTatarnikov: The idea here is to only delete the atlas IF the local already matches the latest version. If the local version is the latest the download machinery doesn't actually change anything. However, if the local version is an older version of the atlas, it's irrelevant whether the force flag is passed. Force or not, the update will occur.

brainglobe_atlasapi/update_atlases.py
IgorTatarnikov: I'll add a line to reset the cache for `local_full_name`.

In theory it should be quite difficult to arrive at this state without intention! If it's during instantiation then the only way for `download` to be called the `local_full_name` has to be `None`, thus line 149 fetches the appropriate `local_full_name`.

Whereas we discard the atlas object used in `update_atlas` once downloaded, as it's only called from the CLI/separately from any existing Atlas objects.

Someone would have to replicate the code inside `update_atlas` outside and then attempt to use the same atlas object. Although, the only way something is deleted is when the local_version matches the remote_version, therefore the local_full_name remains the same following the update.

brainglobe_atlasapi/structure_class.py
IgorTatarnikov: The annoying part is that S3 is always "/" separated, so it has to be mixed throughout.

tests/atlasgen/test_validation.py
IgorTatarnikov: My plan was to raise an issue just before this is merged. It feels a bit weird to raise an issue prior to the code existing in main!

brainglobe_atlasapi/core.py
```diff
@@ -39,18 +65,68 @@ class Atlas:
...
+            structures_path = (
+                self.root_dir
+                / self.metadata["terminology"]["location"][1:]
+                / V2_STRUCTURES_NAME
+            )
+            structures_list = load_structures_from_csv(structures_path)
+            meshes_dir = (
+                self.metadata["annotation_set"]["location"][1:]
+                + "/"
+                + V2_MESHES_DIRECTORY
```
PolarBean: consider using Path instead of string concatenatioin

brainglobe_atlasapi/list_atlases.py
IgorTatarnikov: Good catch, I've added an early return if the directory doesn't exist.

brainglobe_atlasapi/descriptors.py
IgorTatarnikov: Done

brainglobe_atlasapi/core.py
IgorTatarnikov: I didn't want this to raise a warning right away. We should announce this as an API change, then add a warning, and then deprecate. Open to suggestions otherwise.

brainglobe_atlasapi/core.py
IgorTatarnikov: I added some basic `TqdmCallback`s, we can re-add the fancy `Rich` progress bars by adding a custom class.

brainglobe_atlasapi/bg_atlas.py
IgorTatarnikov: Updated the docstring accordingly.

brainglobe_atlasapi/core.py
IgorTatarnikov: This is converted to a Path object on line 129

PolarBean: why not convert it earlier to avoid the concat? or is this another s3 path

PolarBean: maybe we should add s3 to the variable name of s3 paths?

IgorTatarnikov: I tried to stick to using a `remote` prefix where possible, but I can update this to be `s3` instead.

I'm leaning toward creating a helper function that formats the remote path correctly regardless of what's requested.

adamltyson: Agree, we shouldn't bake s3 into the variable names. We will end up moving away from s3 at some point.

tests/atlasapi/test_core_atlas.py
```diff
@@ -183,11 +175,13 @@ def test_meshfile_from_id(atlas):
     atlas : brainglobe_atlasapi.core.Atlas
         The atlas fixture.
     """
-    assert (
-        atlas.meshfile_from_structure("CH")
-        == atlas.root_dir / "meshes/567.obj"
+    mesh_rooth_path = (
```
alessandrofelder: rename to `mesh_root_path` (typo)?

brainglobe_atlasapi/update_atlases.py
```diff
@@ -29,32 +29,35 @@ def update_atlas(atlas_name, force=False, fn_update=None):
...
-            rprint(
-                f"[b][magenta2]brainglobe_atlasapi: {atlas.atlas_name} "
-                "is already updated "
-                f"(version: {_version_str_from_tuple(atlas.local_version)})"
-                "[/b]"
+    is_latest_version = atlas.check_latest_version(print_warning=False)
+
+    if force and is_latest_version:
+        # Delete atlas folder to force update
+        fld = (atlas.brainglobe_dir / atlas.local_full_name).parent
```
alessandrofelder: Nitpicky, opinionated and biased towards long variable names: should `fld` be renamed to `atlas_folder`?

brainglobe_atlasapi/bg_atlas.py
```diff
@@ -1,72 +1,125 @@
...
 
+    Caught by ``_FallbackToLegacyMeta`` to trigger a transparent retry with
+    ``BrainGlobeAtlasLegacy``. Subclasses ``ValueError`` for backward
+    compatibility with callers that catch ``ValueError``.
+    """
 
-def _version_str_from_tuple(version_tuple: Tuple[int, ...]) -> str:
-    return f"{version_tuple[0]}.{version_tuple[1]}"
 
+class _FallbackToLegacyMeta(type):
```
alessandrofelder: Fancy :grin:

brainglobe_atlasapi/bg_atlas.py
```diff
@@ -75,145 +128,270 @@ def __init__(
...
             self.check_latest_version()
 
     @property
-    def local_version(self) -> Optional[Tuple[int, ...]]:
-        """If atlas is local, return actual version of the downloaded files;
-        Else, return none.
+    def local_full_name(self):
         """
-        full_name = self.local_full_name
+        Returns the local full path to the manifest.json file of the atlas.
```
alessandrofelder, line 122: Should this docstring mention that this will return the path to the manifest file either of the latest locally available version or of `self._requested_version` (IIUC)?

brainglobe_atlasapi/bg_atlas.py
```diff
@@ -75,145 +128,270 @@ def __init__(
...
+                path_str.split("/")[-1] for path_str in versions_path
+            ]
+            available_versions.sort(
+                key=lambda v: tuple(int(x) for x in v.split("_")), reverse=True
+            )
 
-        return None
+            self._remote_version = _version_tuple_from_str(
+                available_versions[0].replace("_", ".")
+            )
```
alessandrofelder: Worth splitting this out into a private helper function (we do this twice) with a name like `get_latest_version` to make the intention of the `[0]` and the sorting really explicit?

alessandrofelder (commented):
Small nitpicky comments on the code side.
Generally, I would be happy to approve pending more user testing in the next two days.

(I've done one user test and it was SO fast :tada: thanks, @IgorTatarnikov )

PolarBean:
I might have come across a bug. I've been trying to use this branch in my day to day work and ran into a crash while running brainrender. 

```python
scene = Scene(atlas_name="allen_mouse_25um", check_latest=False, title="Labelled cells")
```
returns 
```
The region root is in the onthology but does not have a corresponding volume in the atlas being used: allen_mouse_25um. Skipping
```
```python

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[9], [line 1](vscode-notebook-cell:?execution_count=9&line=1)
----> [1](vscode-notebook-cell:?execution_count=9&line=1) scene = Scene(atlas_name="allen_mouse_25um", check_latest=False, title="Labelled cells")

File ~/anaconda3/envs/CCFT/lib/python3.14/site-packages/brainrender/scene.py:79, in Scene.__init__(self, root, atlas_name, check_latest, inset, title, screenshots_folder, plotter, title_color)
     72 self.root = self.add_brain_region(
     73     "root",
     74     alpha=settings.ROOT_ALPHA,
     75     color=settings.ROOT_COLOR,
     76     silhouette=bool(root and settings.SHADER_STYLE == "cartoon"),
     77 )
     78 self.atlas.root = self.root  # give atlas access to root
---> [79](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/harryc/github/PyNutil/~/anaconda3/envs/CCFT/lib/python3.14/site-packages/brainrender/scene.py:79) self._root_mesh = self.root.mesh.clone()
     80 if not root:
     81     self.remove(self.root)

AttributeError: 'NoneType' object has no attribute 'mesh'
```
and 
```python
from brainrender.atlas import Atlas
a = Atlas("allen_mouse_25um", check_latest=False)
print(a.get_region("root"))
```
returns
```
The region root is in the onthology but does not have a corresponding volume in the atlas being used: allen_mouse_25um. Skipping
```

This is fixed by switching over to the latest PyPI release

brainglobe_atlasapi/update_atlases.py
IgorTatarnikov: Sounds good! I've changed the name

tests/atlasapi/test_core_atlas.py
IgorTatarnikov: Good catch!

brainglobe_atlasapi/bg_atlas.py
IgorTatarnikov: Updated

brainglobe_atlasapi/bg_atlas.py
IgorTatarnikov: Good idea, I've extracted it into a separate function

alessandrofelder (approved):
I have had a final play, instantiating an atlas and visualising its parts using the `napari` IPython console, using the latest version. It works really nicely:
image

And I can confirm I can access meshes, annotations and template smoothly programmatically.
Approving!

PR #774, opened 2026-03-15
Invalidmesh fix for blue brain barrels

HarshdipSaha:
## Description

**What is this PR**

- [X] Bug fix
- [ ] Other

**Why is this PR needed?**
When we run allen mouse blue brain barrels mesh generation multiple times caching the result,after some runs the max value in mesh becomes so large that is cannot be fitted within the memory limits, leading to crash or unexpected behavior.

**What does this PR do?**

This pr does a simple fix i.e in the first run, when the mesh is scaled for the first time, in cache it also create a _scaled.obj copy, and after that any time the script is ran again, if this _scaled obj exists, it doesn;t multiply by resolution, directly use it.

## References
#763 

Previous code on running and generating meshes multiplied value by resoltion and kept on increasing as the runs increased until overflow.
Now,
```console
Creating meshes ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 9:07:43
Finished mesh extraction in : 547.72 minutes
Finalising atlas
Running atlas validation on allen_mouse_bluebrain_barrels_10um_v1.0
These validation functions have failed:
- catch_missing_mesh_files: Structures with IDs [545] are in the atlas, but don't have a corresponding mesh file.

Run 1 max_abs_vertex = 8825.0024
```
After running the atlas generation again:
```console
Creating meshes ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00
Finished mesh extraction in : 0.0 minutes
Finalising atlas
Running atlas validation on allen_mouse_bluebrain_barrels_10um_v1.0
These validation functions have failed:
- catch_missing_mesh_files: Structures with IDs [545] are in the atlas, but don't have a corresponding mesh file.

Run 2 max_abs_vertex = 8825.0024
```
The vertex values remain stable across runs because _scaled.obj is reused instead of applying scaling repeatedly.
## Is this a breaking change?

no

no
## Checklist:

- [X] The code has been tested locally
- [X]  Tests have been added to cover all new functionality (unit & integration)
- [X]  The code has been formatted with [pre-commit](https://pre-commit.com/)


PolarBean:
Hi @HARSHDIPSAHA this looks good. Could you please package the atlases at 10 and 25 micron resolution and then provide a link to the files. If this is not possible that's fine,

HARSHDIPSAHA:
Hi @PolarBean I have the 10 micron file saved with me.

HARSHDIPSAHA:
https://drive.google.com/file/d/1U59v5Y7xNc8T4zOMzFDB6baxbgnEW-E_/view?usp=sharing

atlas_scripts/allen_mouse_bluebrain_barrels.py
PolarBean: now there is two ways to reach create region mesh. Is it the case that a region file not pre exist AND not be in barrel structure ids? doesnt barrel structure ids cover all non already existing meshes?

PolarBean (changes requested):

atlas_scripts/allen_mouse_bluebrain_barrels.py
HARSHDIPSAHA: This is just used as a fallback in case the server is down or the download crashes.(this logic already exists, you are seeing this function used two times because of extra case of saving the scaled obj).

HARSHDIPSAHA: We can create one single function to handle this cases-precomputed, scaled cache and fallback if this duplication of function is a problem, I honestly think it is fine.

PolarBean: I don't think its good to have nondeterministic functions like this that produce different atlases depending on whether or not the server is up. if the servers down the script will crash anyway given it needs to download the template and annotation

HARSHDIPSAHA: yeah thats a good point ..this non determinism is not a good thing here, I will safely remove the nonbarrel fallback then.

HARSHDIPSAHA: so if the standard mesh is missing we can raise a warning or file not found error?

PolarBean: sure thats good

HARSHDIPSAHA:
@PolarBean thanks for looking into this. I have addressed the changes.

atlas_scripts/allen_mouse_bluebrain_barrels.py
```diff
@@ -328,44 +334,56 @@ def recurse(sub_d):
...
+                    )
+                )
+
+            try:
+                mesh = mio.read(file_name)
+                mesh.points *= resolution
+                mio.write(scaled_file_name, mesh)
+                meshes_dict[node.identifier] = scaled_file_name
+            except mio._exceptions.ReadError:
+                print(f"Mesh file {file_name} not found.")
```
PolarBean, line 368: it could also be corrupt. maybe say mesh file not found or malformed

atlas_scripts/allen_mouse_bluebrain_barrels.py
```diff
@@ -328,44 +334,56 @@ def recurse(sub_d):
 
         # Check if mesh already exists
         file_name = meshes_dir / f"{node.identifier}.obj"
+
+        # this Barrel meshes are generated in voxel units & need scaling once.
```
PolarBean, line 338: These barrel meshes*

atlas_scripts/allen_mouse_bluebrain_barrels.py
PolarBean: If create_region_mesh runs and produces file_name, but mio.read then raises ReadError, the unscaled file_name remains on disk. On the next run, file_name.exists() will be True, so create_region_mesh is skipped but the mio.read will fail again, silently. Consider either deleting file_name on ReadError, or raising rather than printing.

atlas_scripts/allen_mouse_bluebrain_barrels.py
```diff
@@ -328,44 +334,56 @@ def recurse(sub_d):
...
+                meshes_dict[node.identifier] = scaled_file_name
+            except mio._exceptions.ReadError:
+                print(f"Mesh file {file_name} not found.")
+            continue
+
         if file_name.exists():
             meshes_dict[node.identifier] = file_name
             continue
 
         else:
```
PolarBean, line 375: slightly nitpicky but else is redundant when the previous if ends in continue

PolarBean (changes requested):

PolarBean:
Hey @HarshdipSaha, are you able to get back to this or should we close it for the moment? (you are welcome to come back to it)

alessandrofelder:
CLosing as inactive for a while. Reopen if needed.

PR #777, opened 2026-03-17
fix: improve config.py with type  hints, error handling, and missing API

Nityahiray:
## Description

**What is this PR**

- [ ] Bug fix

**Why is this PR needed?**

The 'config.py' module had several small but impactful issues:

1. 'cli_modify_config' had 'key=0' and 'value=0' as defaults - integers where strings are expected, which is a silent bug.
2. 'write_config_value' silently did nothing when passed an unknown key, making debugging very difficult.
3. '_print_config' was printing the '[DEFAULT]'  section, which is a configparser internal artifact and should not appear in user output.
4.  'get_interm_download_dir()' was missing despite 'interm_download_dir' existing in the config - unlike 'brainglobe_dir' which already had 'get_brainglobe_dir()'.
5. No type hints existed anywhere in the module, making it harder to use and maintain.
**What does this PR do?**

## References

No existing issues - these are independently identified improvements.

- Manually tested 'write_config_value' with a valid key (works as before) and an invalid key(now raises 'keyError' with  a clear message)
- Manually tested '_print_config' and confirmed '[DEFAULT]' section no longer appears in output
- Manually tested 'get_interm_download_dir()' returns the correct path
- Manually test 'cli_modify_config' with string key/value (wworks as before) and confirmed the default arg but was present before the fix
- Existing functionality confirmed unchanged - all original return types and behaviors are preserved

## Is this a breaking change?

Yes, one minor breaking change: 'write_config_value' now raises 'KeyError' for unknown keys instead of silently doing nothing. Any code that relied on the silent no-op behavior will need to handle the exception. This is intentional - the previous silent failure was itself a bug.

The new public functions 'get_interm_download_dir()' and 'list_config_keys()' should be mentioned in the API reference if one exists. No separate documentation PR has been created yet - happy to add one if maintainers prefer.

## Checklist:

- [ x] The code has been tested locally
- [x ] The documentation has been updated to reflect any changes
- [x ] The code has been formatted with [pre-commit](https://pre-commit.com/)

### DIFF
--- a/brainglobe_atlasapi/config.py
+++ b/brainglobe_atlasapi/config.py
@@ -129,7 +129,7 @@ def list_content_keys(path: Optional[Path] = None) -> list:
             continue
         keys.extend(sect_dict.keys())
     return keys
-    
+
 def get_brainglobe_dir(): -> Path:
     """Return brainglobe default directory.
 
@@ -182,5 +182,5 @@ def _print_config() -> str:
         lines.append(f"[{sect_name}]")
         for k, val in sect_content.items():
             lines.append(f"\t{k}: {val}")
-    
+
     return "\n".join(lines)
--- a/brainglobe_atlasapi/config.py
+++ b/brainglobe_atlasapi/config.py
@@ -9,6 +9,7 @@ BRAINGLOBE_CONFIG_DIR.
 import configparser
 import os
 from pathlib import Path
+from typing import Optional, Union
 
 import click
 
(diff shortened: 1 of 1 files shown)


brainglobe_atlasapi/config.py
```diff
@@ -29,7 +30,8 @@
 DEFAULT_WORKDIR = Path.home() / "brainglobe_workingdir"
 
 
-def write_default_config(path=None, template=None):
+def write_default_config(path=None, template=None, template: Optional[dict] = None,
```
PolarBean, line 33: This is incorrect, the argument is written twice

PolarBean (changes requested):

PolarBean:
@IgorTatarnikov have tests been disabled for PRs im not sure why theyre not running for these

IgorTatarnikov:
If linting fails the tests don't run. The pre-commit check has to pass first.

PolarBean:
@Nityahiray please make sure the linting passes per Igor's comment

PolarBean:
Hey @Nityahiray Im closing this for now, but if youd like to come back to it please reopen :)

PR #793, opened 2026-04-14
Update wrapup scripts for `brainglobe-atlasapi` V3

IgorTatarnikov:
## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**
Following #769 the wrap up function needed updating to use the new format.

**What does this PR do?**
The wrap up function is updated to write files in the new format. The validation functions are updated to check the new format (partially).

The user can choose to provide a dictionary with the following accepted keys: `["name", "version", "skip_saving", "update_existing", "existing_version"]`.
- If name and version are provided the component is saved according to the new spec
- If `use_existing` is set to `True`, metadata for an existing component with `name` and `version` will be fetched and stored in the working directory for future validation.
- If `update_existing` and `existing_version` are provided, the componet with `name` and `exsiting_version` is fully downloaded. This component is merged with the new data (new resolution is inserted into the OME-Zarr), and the updated component is saved under `name` and `version`.
- If the user doesn't specify extra information about an atlas component, the wrap up functions falls back to using the name of the atlas, e.g. `csl_cat`-space and the atlas version.

e.g.
```
template_info = {
	"name": "csl_cat-template,
	"version": "3.1",
	"update_existing": True,
	"existing_version": "3.0",
}
```

Would download the existing `csl_cat-template` version `3.0` and create a new updated `3.1` version.

### Future improvements
- The `wrapup` function currently expects the `mesh_dict` parameter to contain paths to meshes stored as `.obj`. We could change this to accept any mesh format compatbile with `meshio`, and specifically meshes saved as `neuroglancer` format to avoid re-writing meshes during wrapup.
- The users could construct the dataclasses directly inside of the packaging function. Currently, all dataclasses are constructed inside the wrapup function.

## References
Closes #781 

## Is this a breaking change?
No, added a `print` statement to warn users that `compress` and `cleanup_files` no longer have any function.

This should be offically documented and moved to a deprecation warning.

Yes.

## Checklist:


PolarBean:
My preference in the case of a new resolution becoming available is to 

> Insert this into the existing OME-Zarr as a new resolution level and update the minor version.

Ideally I would like for eventually for the resolution parameter to become irrelevant and for users to request an arbitrary resolution that we provide on the fly via https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.RegularGridInterpolator.html

PolarBean:
for number 2 I would say open a pr which merges into this branch so that its easier to review, but thats just my personal preference

atlas_scripts/csl_cat.py
```diff
@@ -452,8 +454,6 @@ def extract_mesh_from_vtk(working_dir):
         resolution=(RESOLUTION,) * 3,
         orientation=ORIENTATION,
         root_id=ROOT_ID,
-        reference_stack=template_volume,
```
PolarBean, line 455: please reset packaging script files to main unless these changes are intentional

brainglobe_atlasapi/atlas_generation/metadata_utils.py
```diff
@@ -53,10 +60,19 @@ def generate_metadata_dict(
         The version of the atlas.
     shape : tuple of int
         The shape (dimensions) of the atlas volume (e.g., (z, y, x)).
-    additional_references : list of str
+    additional_references : list of dict
```
PolarBean: instead of lists of dict where each dict can be out of sync why not make it a list of dataclasses :)

brainglobe_atlasapi/atlas_generation/metadata_utils.py
```diff
@@ -53,10 +60,19 @@ def generate_metadata_dict(
...
         A list of additional reference links or citations.
     atlas_packager : str
         The name of the person or entity packaging the atlas.
+    coordinate_space_metadata : dict
```
PolarBean: could we make these all dataclasses

brainglobe_atlasapi/atlas_generation/wrapup.py
```diff
@@ -2,107 +2,484 @@
...
+
+        # TODO: parallelise and copy if not scaling or reorienting
+        mio.write(
+            mesh_dest_dir / f"{mesh_id}",
+            mesh,
+            file_format="neuroglancer",
+        )
+
+
+def _save_template_data(
```
PolarBean, line 203: save_template writes to output_dir / descriptors.V2_TEMPLATE_NAME

which looks like 
`brainglobe-atlasapi/templates/test_mouse-template/2_3_2_0/anatomical_template.ome.zarr
`

but then save_template_data tries to load 

`brainglobe-atlasapi/templates/test_mouse-template/2_3_2_0
` which makes ngff_zarr raise group not found

the same pattern exists for _save_annotation_data

brainglobe_atlasapi/atlas_generation/wrapup.py
```diff
@@ -2,107 +2,484 @@
...
+
+        _save_if_not_exists(
+            packaging_data.annotation_stack,
+            dest_dir,
+            annotation_info.metadata["name"],
+            transformations,
+            save_annotation,
+        )
+
+        hemispheres_stub = descriptors.format_hemispheres_stub(
```
PolarBean, line 264: format_hemispheres_stub returns the full component path, including the filename. 
for instance
`annotation-sets/test_mouse-annotation/2_3_2_0/hemispheres.ome.zarr
`
but then later save if not exists calls save_hemispheres which appends the filename again so we are writing

annotation-sets/test_mouse-annotation/2_3_2_0/hemispheres.ome.zarr/hemispheres.ome.zarr

brainglobe_atlasapi/atlas_generation/wrapup.py
```diff
@@ -111,251 +488,221 @@ def wrapup_atlas_from_data(
...
+        coordinate_space_info = {
+            "name": f"{atlas_name}-coordinate-space",
+            "version": atlas_version,
+        }
+
+    additional_template_list = []
+    if additional_references is not None:
+        for ref_tuple in additional_references:
+            ref_metadata, _ = ref_tuple
+            if isinstance(ref_metadata, str):
```
PolarBean, line 646: in the doc string it says additional references is a list of tuples containing a dict or a string and validcomponent metadata. but this only runs with string not dict

PolarBean:
the tests pass here but wrapping up locally fails. This could be a good opportunity to update the tests to catch what is missing. do we wrap up a small but real atlas?

brainglobe_atlasapi/atlas_generation/atlas_packaging_data.py
```diff
@@ -0,0 +1,522 @@
...
+
+
+def _load_stack(
+    stack: ValidComponentData,
+) -> List[npt.NDArray]:
+    if isinstance(stack, (str, Path)):
+        return [tifffile.imread(stack)]
+    elif isinstance(stack, np.ndarray):
+        return [stack]
+    return stack
```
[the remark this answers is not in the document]
PolarBean: @IgorTatarnikov this seems like an issue worth looking into (would lists be provided in that case of multiple additional references?), or would you not expect lists here?

brainglobe_atlasapi/atlas_generation/validate_atlases.py
```diff
@@ -292,19 +320,18 @@ def catch_missing_mesh_files(atlas: BrainGlobeAtlas):
...
+    mesh_path = (
+        Path(atlas_path) / meshes_location / descriptors.V2_MESHES_DIRECTORY
+    )
 
-    ids_from_mesh_files = [
-        int(Path(file).stem)
-        for file in os.listdir(obj_path)
-        if file.endswith(".obj")
-    ]
+    ids_from_mesh_files = [int(Path(f).stem) for f in os.listdir(mesh_path)]
```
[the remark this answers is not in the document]
PolarBean: @IgorTatarnikov will we be wrapping up mesh folders which include non mesh metadata files?

brainglobe_atlasapi/atlas_generation/wrapup.py
PolarBean: @IgorTatarnikov is this not an issue?

brainglobe_atlasapi/atlas_generation/atlas_packaging_data.py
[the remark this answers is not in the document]
IgorTatarnikov: It's valid, there should an extra case to check for List[str | Path] and load them into a List[NDArray]

atlas_scripts/csl_cat.py
IgorTatarnikov: Done, good catch.

brainglobe_atlasapi/atlas_generation/metadata_utils.py
IgorTatarnikov: I'll pass the dataclasses to this function directly, instead of generating the dicts outside and passing them.

brainglobe_atlasapi/atlas_generation/metadata_utils.py
IgorTatarnikov: I'll pass the dataclasses to this function directly, instead of generating the dicts outside and passing them.

brainglobe_atlasapi/atlas_generation/validate_atlases.py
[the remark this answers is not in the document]
IgorTatarnikov: They may in the future, I'll add this check here for now

brainglobe_atlasapi/atlas_generation/wrapup.py
IgorTatarnikov: If the first entry of the tuple is a `dict`, then there's nothing that needs to be done. The `str` case is there for backwards compatibility, or ease of use. If the atlas script author just provides a string name then it's formatted into a `dict` with the minimum entries.

brainglobe_atlasapi/atlas_generation/atlas_packaging_data.py
```diff
@@ -0,0 +1,522 @@
...
+
+    If the component is set to skip_saving, will check if it exists remotely
+    and fetch all metadata files locally.
+    If the component is set to update_existing, will check if the existing
+    version exists remotely and fetch all data and metadata files locally.
+
+    Parameters
+    ----------
+    component_info : ComponentInfo
+        A dictionary containing the component name, version, and a booleans for
```
PolarBean: this isnt a dict but a dataclass

brainglobe_atlasapi/atlas_generation/metadata_utils.py
PolarBean: this says dict but i think its now a dataclass. please review the file for similar instances of this error

PolarBean:
The tests are failing because of the ccfv2 merging before the files being uploaded to AWS. Ignore it and lets get this merged :)

PolarBean (approved):

PR #811, opened 2026-05-20
Fixes docstring that throws error in sphinx autobuild

IgorTatarnikov:
See https://github.com/brainglobe/brainglobe.github.io/issues/494 for context.

This fixes the docstring that lead to the error during sphinx autobuild by adding a new line after the """.


alessandrofelder (approved):
Thanks @IgorTatarnikov !

PR #819, opened 2026-06-01
Add support for 4d annotations

### LINKED ISSUE #820: [BUG] Existing OME-Zarr component metadata fetched from wrong path during packaging
**Describe the bug**
In `brainglobe_atlasapi/atlas_generation/atlas_packaging_data.py`, `check_requested_component` computes `component_stub` by stripping the final path segment (the `.ome.zarr` filename).

When reusing an existing OME-Zarr component (`use_existing` with a `.ome.zarr` `file_name`), the metadata-fetch glob was then appended directly to this *parent* path.

The glob is rooted at the parent directory rather than the specific `.ome.zarr` component, so it matches/pulls metadata to the wrong location, as the files are written into the parent `local_path` instead of the component's `.ome.zarr` directory.

**To Reproduce**
Repackage an atlas while re-using one of the components.

**Expected behaviour**
The component metadata is placed in the correct location.

IgorTatarnikov:

## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**
The [atlas specification](https://atlas-assets.readthedocs.io/en/latest/annotation_set.html) requires that annotations are stored in 2 different formats an `annotations_compressed.ome.zarr` (the current method of storing annotations in BrainGlobe) and a new 4D `annotations.ome.zarr` (num_structures, z, y, x), where each element of the 4D array is a 3D mask for an annotation ID from the terminology. This allows storing annotations fro non-hierarchical regions (e.g. storing a mask for L/R hemisphere), or other annotations that exist outside of the strict hierarchical nature of the current annotations. This also allows support for future probabilistic atlases where each voxel is not assigned to a region in a binary manner.  

**What does this PR do?**
Adds support for 4D annotations in both the fetching and packaging of atlases.

Points the remote to a release candidate endpoint on S3 (this is temporary and will be adjusted prior to V3 release)

## References

Closes #820 

Please explain how any new code has been tested, and how you have ensured that no existing functionality has changed.

## Is this a breaking change?

This is a breaking change for anyone using the current `main` as the file names have changed, making previous atlases invalid.

## Checklist:


atlas_scripts/example_mouse.py
```diff
@@ -179,6 +179,19 @@ def retrieve_or_construct_meshes():
 hemispheres_stack = retrieve_hemisphere_map()
 structures = retrieve_structure_information()
 meshes_dict = retrieve_or_construct_meshes()
+
```
PolarBean, line 182: are these needed for packaging? has the packaging documentation been updated to reference this. 

please include a comment as to why these have been added

brainglobe_atlasapi/atlas_generation/atlas_packaging_data.py
```diff
@@ -81,7 +81,8 @@ def check_requested_component(
     if component_info.use_existing:
         # Add wildcard to fetch all OME-Zarr metadata files
         if component_info.file_name.endswith(".ome.zarr"):
-            remote_path += "/**/*.json"
+            remote_path += f"/{component_info.file_name}/**/*.json"
+            local_path = local_path / component_info.file_name
```
PolarBean, line 85: please create an issue for the bug this fixes and note that this closes that issue

brainglobe_atlasapi/atlas_generation/stacks.py
```diff
@@ -43,6 +43,13 @@
     },
 ]
 
+BG_OME_ZARR_4D_AXES = [
+    {"name": "i", "type": "annotation"},
```
PolarBean: these need to be updated to match the ome zarr axes above

brainglobe_atlasapi/atlas_generation/wrapup.py
```diff
@@ -39,8 +43,15 @@
     ResolutionList,
     ValidComponentData,
 )
+from brainglobe_atlasapi.structure_tree_util import (
+    get_structures_tree,
+    postorder_depth_first_search,
+)
 from brainglobe_atlasapi.utils import atlas_name_from_repr
 
+if TYPE_CHECKING:
```
PolarBean: i think this adds unneccessary complexity for very little savings

brainglobe_atlasapi/atlas_generation/wrapup.py
```diff
@@ -724,7 +913,23 @@ def wrapup_atlas_from_data(
         transformations,
     )
 
-    shapes = [image.data.shape for image in template_multiscale.images]
+    shapes = {}
+
+    for resolution in packaging_data.resolution:
+        # Find the closest matching resolution in the template multiscale
```
PolarBean, line 956: can we instead find the closest matching higher resolution so that we always downsample instead of upsampling where possible?

PolarBean: can we please instead use the physical size and then get a new shape based off the specified voxel size

brainglobe_atlasapi/bg_atlas.py
```diff
@@ -310,6 +311,27 @@ def download(self):
...
+                        + f"/{V3_ANNOTATION_MASKS_NAME}/**/*.json"
+                    )
+                    remote_masks_metadata = remote_url_s3.format(
+                        masks_metadata_glob
+                    )
+                    self.fs.get(
+                        remote_masks_metadata,
+                        str(local_annotation_path / V3_ANNOTATION_MASKS_NAME),
+                        callback=TqdmCallback(),
+                    )
```
PolarBean, line 328: could we have a helper function that we reuse for fetching metadata jsons

PolarBean (approved):

PolarBean:
all looks good! please repackage the test atlases and ensure all tests are passing before merging

atlas_scripts/example_mouse.py
IgorTatarnikov: Added a small comment, further documentation will come in the release notes/blog post related to `brainglobe-atlasapi` V3.

brainglobe_atlasapi/atlas_generation/stacks.py
IgorTatarnikov: Good catch, done

brainglobe_atlasapi/atlas_generation/atlas_packaging_data.py
IgorTatarnikov: Documented in https://github.com/brainglobe/brainglobe-atlasapi/issues/820

brainglobe_atlasapi/atlas_generation/wrapup.py
```diff
@@ -382,21 +388,40 @@ def _compute_4d_masks_for_scale(
...
+    )
+    for k, v in mapping.items():
+        typed_dict[types.uint32(k)] = types.uint32(v)
+
     z, y, x = annotation_scale.shape
     masks = zarr.open_array(
         scratch_path,
         mode="w",
         shape=(n_structures, z, y, x),
-        chunks=(1, z, y, x),
```
PolarBean: why is this a fixed shape?

brainglobe_atlasapi/core.py
```diff
@@ -100,6 +104,22 @@ def __init__(self, path):
...
+        self._annotation_masks_pyramid_level = None
+        masks_path = self._annotation_masks_path
+        if masks_path.exists():
+            masks_multiscale = nz.from_ngff_zarr(masks_path)
+            self._annotation_masks_pyramid_level = _determine_pyramid_level(
+                masks_multiscale, self.resolution
+            )
+            root = zarr.open_group(str(masks_path), mode="r")
+            raw_mapping = root[V3_ANNOTATION_MAP_NAME][:]
+            if raw_mapping is not None:
```
PolarBean: if raw_mapping is a numpy array this is never None and therefore likely not doing what you expect

pyproject.toml
```diff
@@ -24,7 +24,9 @@ dependencies = [
     "fsspec",
     "meshio",
     "ngff-zarr",
+    "numba",
     "numpy",
+    "ome-zarr<0.17.0",
```
PolarBean: this is also an unpinned dependancy in atlas gen. consider keeping only one

brainglobe_atlasapi/descriptors.py
```diff
@@ -8,7 +8,7 @@
 
 # Base url of the gin repository:
 remote_url_base = "https://gin.g-node.org/brainglobe/atlases/raw/master/{}"
-remote_url_s3 = "s3://brainglobe/atlas/{}"
+remote_url_s3 = "s3://brainglobe/atlas-rc2/{}"
 remote_url_s3_http = "https://brainglobe.s3.us-west-2.amazonaws.com/atlas/{}"
```
PolarBean: should the http link also point to rc2? is it intentional to have them differ?

pyproject.toml
```diff
@@ -24,7 +24,9 @@ dependencies = [
     "fsspec",
     "meshio",
     "ngff-zarr",
+    "numba",
```
PolarBean: seems like numba is only used in wrapup/atlasgen.

brainglobe_atlasapi/core.py
```diff
@@ -100,6 +104,22 @@ def __init__(self, path):
...
+        self._annotation_mapping = None
+        self._annotation_masks_pyramid_level = None
+        masks_path = self._annotation_masks_path
+        if masks_path.exists():
+            masks_multiscale = nz.from_ngff_zarr(masks_path)
+            self._annotation_masks_pyramid_level = _determine_pyramid_level(
+                masks_multiscale, self.resolution
+            )
+            root = zarr.open_group(str(masks_path), mode="r")
+            raw_mapping = root[V3_ANNOTATION_MAP_NAME][:]
```
PolarBean, line 116: V3_ANNOTATION_MAP_NAME what happens if this isnt a key in root? judging from this and the below line you likely want something like the old "root.attrs.get("annotation_mapping", None) "

PolarBean (approved):

brainglobe_atlasapi/atlas_generation/wrapup.py
PolarBean: in the new file its now fixed, just checking why

brainglobe_atlasapi/core.py
IgorTatarnikov: This is a remnant of the previous method of storing annotation values. I'll remove the None check.

brainglobe_atlasapi/core.py
IgorTatarnikov: This shouldn't happen, the files that had the annotation_values were never distributed and therefore shouldn't have been on anyone's machine but mine!

brainglobe_atlasapi/descriptors.py
IgorTatarnikov: This should also point to rc2, good catch

pyproject.toml
IgorTatarnikov: Good catch, best to move it there

pyproject.toml
IgorTatarnikov: `ome-zarr` is only needed for atlasgen, moving the dependency and pin there.

IgorTatarnikov (commented):
Thanks for the quick review, I've addressed the comments!

brainglobe_atlasapi/atlas_generation/wrapup.py
IgorTatarnikov: I went back and forth on whether to store an entire mask in one chunk, or split the mask into a few larger chunks. (256, 256, 256) is an in between, each chunk is 16 MB uncompressed with each atlas having at most 64 chunks, roughly a (1024, 1024, 1024) atlas. The masks are downloaded lazily so this shouldn't pollute the directories of users unless they want to look at the mask of every single brain region!

IgorTatarnikov: Happy to go back to storing each mask in it's own chunk as they'll typically be used in entirety, rather than viewed in a lazy manner.

PR #848, opened 2026-06-29
Check HTTP status in retrieve_over_http before writing to disk

### LINKED ISSUE #847: [BUG] retrieve_over_http writes error pages to disk without checking HTTP status → misleading "not a gzip file" ReadError
**Describe the bug**

`utils.retrieve_over_http()` streams `requests.get(url, ...)` straight to disk without ever checking the response status code:

```python
# brainglobe_atlasapi/utils.py
response = requests.get(url, stream=True)
...
with open(output_file_path, "wb") as fout:
    for chunk in response.iter_content(chunk_size=CHUNK_SIZE):
        fout.write(chunk)
```

When the server returns a 403/404/503 (GIN IP blocking, transient outage, a wrong/missing version filename, etc.), the HTML/XML error body is written out *as if it were the atlas archive*. The failure then surfaces one step later in `bg_atlas.download_extract_file()` as a confusing, misdirecting error:

```
ReadError: file could not be opened successfully:
- method gz: ReadError('not a gzip file')
- method bz2: ReadError('not a bzip2 file')
- method xz: ReadError('not an lzma file')
- method tar: ReadError('truncated header')
```

This makes a plain network/HTTP problem look like data corruption, and users have no indication that the real cause is a 403/404/503.

**To Reproduce**

Point `BrainGlobeAtlas` at any atlas/version whose tarball URL doesn't resolve to a 200 (e.g. a version that exists in one catalog but not on GIN, or a download from an IP that GIN 403s — e.g. some Google Colab runtimes). Instead of a clear HTTP error you get the `not a gzip file` ReadError above.

**Expected behaviour**

`retrieve_over_http` should validate the response before writing — e.g. `response.raise_for_status()` (or check `status_code == 200` / `Content-Type`) and surface a clear error like `Failed to download <url>: HTTP 403`, removing any partial file. `conf_from_url()` already checks `status_code`; the data-download path should do the same.

**Suggested fix**

```python
response = requests.get(url, stream=True)
response.raise_for_status()   # fail clearly on 4xx/5xx instead of saving an error page
```

**Additional context**

Hit while running a tutorial notebook on Colab (dandi/example-notebooks#176). A status check would have immediately shown the underlying HTTP 403/404 instead of a tarfile ReadError. Possibly related: #728 (GIN 503s in CI), #433 (download failures).

**Computer used:** brainglobe-atlasapi 2.3.1 (defect also present on the current default branch).

bendichter:
## Description

`utils.retrieve_over_http()` streamed `requests.get(url, ...)` straight to disk without ever checking the response status code. When the server returned a 403/404/503 (GIN IP blocking, transient outage, a wrong/missing version filename, etc.), the HTML/XML error body was written out *as if it were the atlas archive*. The failure then surfaced one step later in `download_extract_file()` as a confusing `ReadError: not a gzip file`, making a plain network/HTTP problem look like data corruption.

## Changes

- Add `response.raise_for_status()` immediately after `requests.get()`, so a 4xx/5xx fails before any error page is written to disk.
- Catch `requests.exceptions.HTTPError`, remove the partial file, and re-raise with a clear `Failed to download file from <url>: HTTP <status>` message.
- Add `test_retrieve_over_http_HTTPError` covering the 403 case (clear error raised, no partial file left behind).

This mirrors the status check that `conf_from_url()` already does for the config-download path.

Closes #847

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/brainglobe_atlasapi/utils.py
+++ b/brainglobe_atlasapi/utils.py
@@ -258,6 +258,9 @@ def retrieve_over_http(
 
     try:
         response = requests.get(url, stream=True)
+        # Fail clearly on 4xx/5xx instead of saving an error page to disk,
+        # which would later surface as a misleading tarfile ReadError.
+        response.raise_for_status()
         with progress:
             tot = int(response.headers.get("content-length", 0))
 
@@ -286,6 +289,12 @@ def retrieve_over_http(
                         # update handler with completed and total bytes
                         fn_update(completed, tot)
 
+    except requests.exceptions.HTTPError as e:
+        output_file_path.unlink(missing_ok=True)
+        status = e.response.status_code if e.response is not None else "?"
+        raise requests.exceptions.HTTPError(
+            f"Failed to download file from {url}: HTTP {status}"
+        ) from e
     except requests.exceptions.ConnectionError:
         output_file_path.unlink(missing_ok=True)
         raise requests.exceptions.ConnectionError(
--- a/tests/atlasapi/test_utils.py
+++ b/tests/atlasapi/test_utils.py
@@ -395,6 +395,34 @@ def test_retrieve_over_http_ConnectionError(tmp_path):
             )
 
 
+def test_retrieve_over_http_HTTPError(tmp_path):
+    """A non-200 response should raise a clear HTTPError and not leave a file.
+
+    Parameters
+    ----------
+    tmp_path : pathlib.Path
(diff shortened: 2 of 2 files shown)


PolarBean:
we used http requests to get atlases previously, but in v3 we use s3 so this is no longer needed. Thanks all the same!

PR #869, opened 2026-07-23
Add tqdm dependency to pyproject.toml

### LINKED ISSUE #868: [BUG] tqdm is a dependancy which is not listed pyproject.toml which causes the api to crash in fresh environments in v3

PolarBean:
closes #868 

reproduced on two machines that this was crashing v3 installs in fresh environments.


IgorTatarnikov (commented):
I agree with the bot! We should remove `tqdm` from the `atlasgen` extras and move it to the core dependencies with the pinned version.

IgorTatarnikov (approved):
LGTM 🎉 Thanks!

PR #872, opened 2026-07-23
switch reference to template in instructions

PolarBean:
we now want users to use template rather than reference


IgorTatarnikov (approved):
LGTM 🚀 

Thanks!

PR #873, opened 2026-07-23
Fix typo in README regarding brain region access

PolarBean:
fix a typo


IgorTatarnikov (approved):
🚀 

Thanks @PolarBean!

PR #879, opened 2026-07-27
Refactor validation functions (part 1 of #782)

a-shahrabi:
## Description

**What is this PR**

- [ ] Bug fix
- [x] Other (refactor)

**Why is this PR needed?**

First of three PRs for #782. The validation functions need updating for the v2 atlas format, but there's some pre-existing duplication and inefficiency in `validate_atlases.py` that's easier to clear out first. Splitting it this way keeps the format-migration diffs small and reviewable.

This PR is a **pure refactor with no behaviour changes**. No validation logic is altered, and the xfail markers in `tests/atlasgen/test_validation.py` are untouched.

**What does this PR do?**

- `validate_atlas` constructed a new `BrainGlobeAtlas` for every validation function in the loop. It now constructs one instance after `update_atlas` and reuses it. The separate bare `BrainGlobeAtlas(atlas_name)` call above is deliberately kept, since it triggers the download so the atlas appears in `get_atlases_lastversions()`, and I've added a comment saying so.
- Extracted the duplicated structure-ID/mesh-ID lookup shared by `catch_missing_mesh_files` and `catch_missing_structures` into `_get_structure_and_mesh_ids`. Both functions keep their own comparison direction and assertion messages.
- `validate_atlas` caught only `AssertionError`, so a `ValueError` from `int(f.stem)` or an S3/OS error would abort the entire run and lose all results collected so far. Non-assertion exceptions are now recorded with a distinct `"Error"` status, keeping "atlas is invalid" distinguishable from "the check itself failed".

One thing I looked at and deliberately left alone: `report_validation_results` expects a flat `{func_name: "Pass"/"Fail: msg"}` dict, while `validate_atlas` returns `{atlas_name: [(func_name, error, status)]}`. These are incompatible, but `wrapup.py` is the only caller and it builds the flat shape itself, so there's no live bug. The underlying issue is that there are two independent implementations of "run all validators and collect results". I've raised that as a question on #782 rather than changing it here.

## References

Part 1 of 3 for #782. Follows on from #769.

Full test suite run locally before and after: 332 passed, 5 xfailed, unchanged. The 5 xfails are the validation tests awaiting the v2 format work in the later PRs.

No behaviour was intentionally changed, so no test changes were needed. I reviewed each diff hunk to confirm the extracted logic and the assertion messages are identical to before.

## Is this a breaking change?

No.

No. There are no public API or behaviour changes, and `_get_structure_and_mesh_ids` is private.

## Checklist:

- [ ] Tests have been added to cover all new functionality (unit & integration). Not applicable, no new functionality.
- [ ] The documentation has been updated to reflect any changes. Not applicable.
- [x] The code has been formatted with [[pre-commit](https://pre-commit.com/)](https://pre-commit.com/)

### DIFF
--- a/brainglobe_atlasapi/atlas_generation/validate_atlases.py
+++ b/brainglobe_atlasapi/atlas_generation/validate_atlases.py
@@ -693,16 +693,18 @@ def validate_atlas(atlas_name, version, validation_functions):
         `error_message` is None if the check passes.
     """
     print(atlas_name, version)
+    # Triggers download so the atlas appears in get_atlases_lastversions()
     BrainGlobeAtlas(atlas_name)
     updated = get_atlases_lastversions()[atlas_name]["updated"]
     if not updated:
         update_atlas(atlas_name)
 
+    atlas = BrainGlobeAtlas(atlas_name)
     validation_results = {atlas_name: []}
 
-    for i, validation_function in enumerate(validation_functions):
+    for validation_function in validation_functions:
         try:
-            validation_function(BrainGlobeAtlas(atlas_name))
+            validation_function(atlas)
             validation_results[atlas_name].append(
                 (validation_function.__name__, None, str("Pass"))
             )
--- a/brainglobe_atlasapi/atlas_generation/validate_atlases.py
+++ b/brainglobe_atlasapi/atlas_generation/validate_atlases.py
@@ -297,6 +297,39 @@ def validate_additional_references(atlas: BrainGlobeAtlas):
     return True
 
 
+def _get_structure_and_mesh_ids(atlas: BrainGlobeAtlas):
+    """Get structure IDs from the atlas and mesh IDs from the meshes folder.
+
+    Parameters
+    ----------
+    atlas : BrainGlobeAtlas
+        The BrainGlobeAtlas object to inspect.
+
+    Returns
+    -------
+    tuple of (list of int, list of int)
+        The structure IDs from `atlas.structures` and the mesh IDs found
+        in the atlas's `meshes` directory.
+    """
+    ids_from_bg_atlas_api = list(atlas.structures.keys())
+
+    atlas_path = atlas.root_dir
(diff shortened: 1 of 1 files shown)


brainglobe_atlasapi/atlas_generation/validate_atlases.py
```diff
@@ -297,25 +297,19 @@ def validate_additional_references(atlas: BrainGlobeAtlas):
     return True
 
 
-def catch_missing_mesh_files(atlas: BrainGlobeAtlas):
-    """Check if all structures in the atlas have a corresponding mesh file.
+def _get_structure_and_mesh_ids(atlas: BrainGlobeAtlas):
```
IgorTatarnikov:
```suggestion
def _get_structure_and_mesh_ids(atlas: BrainGlobeAtlas) -> tuple[list[int], list[int]]:
```

IgorTatarnikov (commented):
Just one small thing, could you type hints for the returns in the validation functions? 

Apart from that this is good to go!

brainglobe_atlasapi/atlas_generation/validate_atlases.py
a-shahrabi: I added return hints to the validation functions and the _get_structure_and_mesh_ids helper. I left validate_atlas and get_all_validation_functions unannotated since they're not validators, but happy to add those too if you'd prefer.
Whenever you get a chance, I left two follow-ups on #782 about the zarr metadata checks and whether the format_*_stub descriptors should be the schema source for validation. PR 2 is blocked on the second one.

a-shahrabi:
@IgorTatarnikov, no rush, just checking in on this one. The type hints are in and it's been green since. Happy to rebase if it's gone stale.

Also, PR 2 (#944) is also ready, though CI there is blocked on the nmt_arm_sym_macaque_250um / AtlasName mismatch I flagged, which affects any PR in the repo and I've got answers to most of the PR 3 questions from my digging on #782 whenever you have a moment.

PR #902, opened 2026-08-03
add optional license information to atlases

PolarBean:
This adds an optional metadata field to atlas wrapup for inclusion of a license. The most useful way to do this i thought would be to have a license short name, however many atlases use a custom atlas. as a compromise I propose this as the best way to integrate licensing. 

```python
        "license": {
            "license": "CC-BY-4.0",
            "link_to_license": (
                "https://creativecommons.org/licenses/by/4.0/"
            ),
        }
```

Ive also made this directly accessible via atlas.license.

In cases where its not set atlas.license will return None. This is distinct from 

```python
        "license": {
            "license": "unlicensed"
            "link_to_license": (
                None
            ),
        }
```

In cases of a custom license we can say 

```python
        "license": {
            "license": "custom"
            "link_to_license": (
                "https://alleninstitute.org/legal/terms-of-use"
            ),
        }
```

### DIFF
--- a/brainglobe_atlasapi/atlas_generation/atlas_packaging_data.py
+++ b/brainglobe_atlasapi/atlas_generation/atlas_packaging_data.py
@@ -382,8 +382,11 @@ class AtlasPackagingData:
     atlas_packager : str, optional
         Credit for those responsible for converting the atlas into the
         BrainGlobe format.
-    license : Dict[str, str], optional
-        License metadata with ``license`` and ``link_to_license`` fields.
+    license : Dict[str, str | None] or None, optional
+        License metadata with ``license`` and ``link_to_license`` fields. The
+        ``link_to_license`` value may be ``None`` for an explicitly unlicensed
+        atlas. The entire field may also be ``None`` when license metadata is
+        not provided.
     hemispheres_stack : ValidComponentData, optional
         Hemisphere stack for the atlas. If None, atlas is assumed symmetric.
     additional_references : List[Tuple[TemplateInfo, ValidComponentData]], optional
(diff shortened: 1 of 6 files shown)


PolarBean:
something to discuss here is that openminds provides a license field we could borrow from if we feel this one is underspecified. It is [here](https://github.com/openMetadataInitiative/openMINDS_core/blob/v6/schemas/data/license.schema.tpl.json) and looks like the following. 


```python
{
  "_type": "core:License",
  "required": [
    "fullName",
    "legalCode",
    "shortName"
  ],
  "properties": {
    "fullName": {
      "type": "string",
      "_instruction": "Enter the full name of this license."
    },
    "legalCode": {
      "type": "string",
      "_formats": [
        "iri"
      ],
      "_instruction": "Enter the internationalized resource identifier (IRI) to the legal code of this license."
    },
    "shortName": {
      "type": "string",
      "_instruction": "Enter a short name (or alias) for this license that could be used as a shortened display title (e.g., for web services with too little space to display the full name)."
    },
    "webpage": {
      "type": "array",
      "minItems": 1,
      "uniqueItems": true,
      "_instruction": "Enter the internationalized resource identifiers (IRIs) to webpages related to this license (e.g., a homepage).",
      "items": {
        "type": "string",
        "_formats": [
          "iri"
        ]
      }
    }
  }
}
```

I think it is overkill for our needs and will make packaging more complex but perhaps something in between is good (Im not against keeping the simple implementation proposed in this PR).

adamltyson:
As discussed offline, we should try to adopt the AIND metadata spec for this (and other things), and expose through `atlas.metadata`

PR #904, opened 2026-08-04
Speed up test suite

IgorTatarnikov:

## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**
Local atlas existence was checked before every test in `atlasgen`. Due to a change in how the atlases are stored on disk, the existence check was always `False` leading to a `BrainGlobeAtlas` instantiation with `check_latest=True`. This has quite a few network calls, even if the atlas is available on disk. This lead to a slow down in the overall test suite.

**What does this PR do?**
Updates the existence check, and moves that fixture to be session scoped.

The `mock_brainglobe-user-folders` fixture also needed to be fixture scoped to make sure it runs before the local atlas check. This needs a manual `MonkeyPatch` as the `monkeypatch` fixture can't be used at a session scope.

All tests pass locally, test time has gone from ~5-10 mins to 1.5 mins. I verified the `mock_brainglobe_user_folders` fixture still fires by debugging a random test and ensuring that `Path.home()` returned the correct value.

## Is this a breaking change?
No

No

## Checklist:

### DIFF
--- a/tests/atlasgen/conftest.py
+++ b/tests/atlasgen/conftest.py
@@ -4,14 +4,16 @@ Provide commonly used fixtures and setup for testing `brainglobe_atlasapi`
 functionality.
 """
 
-from pathlib import Path
-
 import pytest
 
 from brainglobe_atlasapi import BrainGlobeAtlas
+from brainglobe_atlasapi.list_atlases import (
+    get_downloaded_atlases,
+    get_local_atlas_version,
+)
 
 
-@pytest.fixture(autouse=True)
+@pytest.fixture(autouse=True, scope="session")
 def setup_preexisting_local_atlases():
     """Set up all tests to have three downloaded atlases in the test user data.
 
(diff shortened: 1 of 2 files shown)


tests/atlasgen/conftest.py
```diff
@@ -4,29 +4,33 @@
...
     """Set up all tests to have three downloaded atlases in the test user data.
 
     Automatically downloads and sets up predefined atlases for testing
     purposes, ensuring they are available locally before test execution.
     """
     preexisting_atlases = [
-        ("example_mouse_100um", "v1.2"),
-        ("allen_mouse_100um", "v1.2"),
-        ("kim_dev_mouse_e11-5_mri-adc_31.5um", "v1.3"),
+        ("example_mouse_100um", "3.0"),
```
PolarBean: please unpin these so that they dont redownload once the atlases update.

tests/conftest.py
```diff
@@ -25,11 +25,19 @@ def mock_brainglobe_user_folders(monkeypatch):
...
+    download atlases (e.g. `tests/atlasgen/conftest.py`); a function-scoped
+    mock would run too late and those would use the real home directory.
+
     Note
     ----
     GitHub Actions workflow will test with default user folders.
     """
-    if not os.getenv("GITHUB_ACTIONS"):
+    if os.getenv("GITHUB_ACTIONS"):
+        yield
```
PolarBean: This yields seems unnecessary

PolarBean (approved):
LGTM

tests/conftest.py
PolarBean: i take it back

PR #907, opened 2026-08-05
Report atlas download progress to fn_update

### LINKED ISSUE #874: [BUG] `fn_update` is never called by `BrainGlobeAtlas`
**Describe the bug**
We accept a `fn_update` as an argument in the `BrainGlobeAtlas` constructor. This is to facilitate updating of external progress bars. This is never called by `BrainGlobeAtlas`. 

**Expected behaviour**
External progress bars are updated on install and component fetching steps.

adityasingh2400:
Fixes #874.

`BrainGlobeAtlas` takes an `fn_update` handler so that external progress bars can follow an install, and the docstring says it receives completed and total bytes. The V3 download path passes a plain `TqdmCallback()` to all nine `fs.get` calls and never touches the handler, so downstream progress bars never move. brainrender-napari passes one in `AtlasManagerView._download_atlas_install` and `_download_atlas_update`, and it does nothing today.

This adds a callback that keeps the tqdm bar and attaches `fn_update` to the per file callbacks fsspec creates through `branched`. fsspec drives the callback given to `get` with a file count and drives the branched callback with a byte count, so attaching the handler to the branched one keeps the documented byte semantics. When no handler is given the behaviour is unchanged.

Tests cover the callback against an fsspec memory filesystem for both a single file and a recursive transfer, plus an end to end check that a real `BrainGlobeAtlas(..., fn_update=...)` install calls the handler.

Reverting `bg_atlas.py` alone gives 4 failed and 10 passed, with `AssertionError: fn_update was not called during the atlas download`. After, the full `tests/atlasapi` suite is green at 119 passed.

Disclosure: this change was prepared with AI assistance. I have reviewed and tested it.

### DIFF
--- a/brainglobe_atlasapi/bg_atlas.py
+++ b/brainglobe_atlasapi/bg_atlas.py
@@ -7,7 +7,7 @@ from pathlib import Path
 from typing import List, Optional, Tuple, Union
 
 import s3fs
-from fsspec.callbacks import TqdmCallback
+from fsspec.callbacks import Callback, TqdmCallback
 from rich import print as rprint
 from rich.console import Console
 
@@ -38,6 +38,58 @@ def _version_tuple_from_str(version_str):
 
 def _version_str_from_tuple(version_tuple: Tuple[int, ...]) -> str:
     return "_".join(str(num) for num in version_tuple)
+
+
+class _ProgressCallback(TqdmCallback):
+    """Show a tqdm bar and report progress to an external handler.
+
+    ``fsspec`` drives the callback passed to ``get`` with a file count, and
+    asks it for a per-file callback through ``branched`` which is driven with
+    a byte count. ``fn_update`` is documented to take completed and total
+    bytes, so it is attached to the branched callbacks.
+
+    Parameters
+    ----------
+    fn_update : Callable
+        Handler called as ``fn_update(completed, total)`` with the bytes
+        transferred so far and the total size of the file being fetched.
+    """
+
+    def __init__(self, fn_update: Callable[[int, int], None], **kwargs):
+        super().__init__(**kwargs)
+        self.fn_update = fn_update
+
+    def branched(self, path_1, path_2, **kwargs):
+        """Return a per-file callback that reports bytes to ``fn_update``."""
+        kwargs["callback"] = Callback(
+            hooks={
+                "fn_update": lambda size, value, **_: self.fn_update(
+                    value, size or 0
+                )
+            }
+        )
+        return super().branched(path_1, path_2, **kwargs)
+
+
+def _download_callback(
+    fn_update: Optional[Callable[[int, int], None]],
+) -> TqdmCallback:
+    """Build the fsspec callback used for a single atlas download step.
+
+    Parameters
+    ----------
+    fn_update : Callable, optional
+        Handler to report download progress to. If None, only the tqdm
+        progress bar is shown.
+
+    Returns
+    -------
+    TqdmCallback
+        The callback to pass to ``fsspec``.
+    """
+    if fn_update is None:
+        return TqdmCallback()
+    return _ProgressCallback(fn_update)
 
 
 class BrainGlobeAtlas(core.Atlas):
@@ -246,7 +298,11 @@ class BrainGlobeAtlas(core.Atlas):
             f"Downloading {self.atlas_name} atlas "
             f"v{remote_version_str.replace('_', '.')} manifest:"
         )
-        self.fs.get(remote_path, local_path, callback=TqdmCallback())
+        self.fs.get(
+            remote_path,
+            local_path,
+            callback=_download_callback(self.fn_update),
+        )
         self.metadata = read_json(local_path)
 
         try:
@@ -265,7 +321,7 @@ class BrainGlobeAtlas(core.Atlas):
                     remote_terminology_path,
                     local_terminology_path,
                     recursive=True,
-                    callback=TqdmCallback(),
+                    callback=_download_callback(self.fn_update),
                 )
 
             # Download coordinate space files
@@ -285,7 +341,7 @@ class BrainGlobeAtlas(core.Atlas):
                     remote_coordspace_path,
                     local_coordspace_path,
                     recursive=True,
-                    callback=TqdmCallback(),
+                    callback=_download_callback(self.fn_update),
                 )
 
             # Download annotation metadata files
@@ -307,7 +363,7 @@ class BrainGlobeAtlas(core.Atlas):
                 self.fs.get(
                     remote_root_metadata_path,
                     local_annotation_path / V3_ANNOTATION_NAME,
-                    callback=TqdmCallback(),
+                    callback=_download_callback(self.fn_update),
                 )
                 mesh_path = local_annotation_path / V3_MESHES_DIRECTORY
                 mesh_path.mkdir(parents=True, exist_ok=True)
@@ -324,7 +380,7 @@ class BrainGlobeAtlas(core.Atlas):
                     self.fs.get(
                         remote_masks_metadata,
                         str(local_annotation_path / V3_ANNOTATION_MASKS_NAME),
-                        callback=TqdmCallback(),
+                        callback=_download_callback(self.fn_update),
                     )
                     masks_annotation_values_path = (
                         annotation_location + f"/{V3_ANNOTATION_MASKS_NAME}"
@@ -336,7 +392,7 @@ class BrainGlobeAtlas(core.Atlas):
                     self.fs.get(
                         remote_masks_annotation_values_path,
                         str(local_annotation_path / V3_ANNOTATION_MASKS_NAME),
-                        callback=TqdmCallback(),
+                        callback=_download_callback(self.fn_update),
                         recursive=True,
                     )
                 except FileNotFoundError as e:
@@ -357,7 +413,7 @@ class BrainGlobeAtlas(core.Atlas):
                     self.fs.get(
                         remote_root_hemisphere_path,
                         local_annotation_path / V3_HEMISPHERES_NAME,
-                        callback=TqdmCallback(),
+                        callback=_download_callback(self.fn_update),
                     )
 
             # Download template metadata files
@@ -380,7 +436,7 @@ class BrainGlobeAtlas(core.Atlas):
                 self.fs.get(
                     remote_root_metadata_path,
                     local_template_path / V3_TEMPLATE_NAME,
-                    callback=TqdmCallback(),
+                    callback=_download_callback(self.fn_update),
                 )
 
             additional_reference_names = self.metadata.get(
@@ -405,7 +461,7 @@ class BrainGlobeAtlas(core.Atlas):
                     self.fs.get(
                         remote_root_metadata_path,
                         local_template_path / V3_TEMPLATE_NAME,
-                        callback=TqdmCallback(),
+                        callback=_download_callback(self.fn_update),
                     )
             # Reset local_full_name to ensure it is updated with new location
             self._local_full_name = None
--- a/brainglobe_atlasapi/bg_atlas.py
+++ b/brainglobe_atlasapi/bg_atlas.py
@@ -12,7 +12,7 @@ from rich.console import Console
 
 from brainglobe_atlasapi import config, core
 from brainglobe_atlasapi.atlas_name import AtlasName
-from brainglobe_atlasapi.core import _download_callback
+from brainglobe_atlasapi.callback import _download_callback
 from brainglobe_atlasapi.descriptors import (
     V3_ANNOTATION_MAP_NAME,
     V3_ANNOTATION_MASKS_NAME,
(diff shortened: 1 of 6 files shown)


IgorTatarnikov:
Hello, this is a great start! I believe there are some progress bars in `brainglobe_atlasapi/core.py` as well. I think this solution should be generalised to apply to both!

adityasingh2400:
Good catch, done in 000957c. You are right that the download was only half the story.

`core.Atlas` fetches the template, the annotation, the hemispheres, and the additional references lazily from the same store, each with its own bare `TqdmCallback()`, five call sites in total. So a caller who passed `fn_update` got byte progress for the initial download and then silence for every pyramid level fetched afterwards, which is probably the part they were waiting on.

What changed:

`_ProgressCallback` and `_download_callback` moved from `bg_atlas.py` into `core.py`. `core` is the lower level module and `bg_atlas` already imports it, so this direction avoids a circular import and lets both use the same helper rather than duplicating it.

`core.Atlas.__init__` takes an optional `fn_update` and stores it, and all five lazy fetches now use `_download_callback(self.fn_update)`. A `core.Atlas` constructed directly can report progress too, not only a `BrainGlobeAtlas`.

`BrainGlobeAtlas` passes its handler down through `super().__init__` instead of keeping it to itself. It still sets `self.fn_update` before that call because `download()` runs first and needs it.

Behavior with no handler is unchanged: `_download_callback(None)` returns a plain `TqdmCallback`, so the existing bars look exactly as before.

Verified with `ruff check` and `ruff format --check` clean across the package and the touched test, and both modules compile. I added a test in `test_core_atlas.py` covering that a handler produces a `_ProgressCallback` carrying it and that no handler produces a plain `TqdmCallback`.

One thing I did not do: I left the five sites reporting per-file byte counts rather than aggregating them into a single overall figure across the whole atlas. That seemed like a separate decision about what the number should mean, so say the word if you would rather have the aggregate.

brainglobe_atlasapi/core.py
```diff
@@ -66,7 +120,11 @@ class Atlas:
     left_hemisphere_value = 1
     right_hemisphere_value = 2
 
-    def __init__(self, path):
+    def __init__(self, path, fn_update: Optional[Callable] = None):
+        # Progress handler shared by every lazy fetch below, so a caller that
+        # passes one sees byte progress for the pyramid levels too and not only
+        # for the initial download.
```
IgorTatarnikov: This comment doesn't make sense here, it should be removed
```suggestion
```

brainglobe_atlasapi/core.py
```diff
@@ -54,6 +56,58 @@ def _determine_pyramid_level(
...
+        progress bar is shown.
+
+    Returns
+    -------
+    TqdmCallback
+        The callback to pass to ``fsspec``.
+    """
+    if fn_update is None:
+        return TqdmCallback()
+    return _ProgressCallback(fn_update)
```
IgorTatarnikov: I think this is best placed in a separate `callback` module.

IgorTatarnikov (changes requested):
Hi @adityasingh2400, thanks for making the changes. Just a few more comments and then it'll be good to go.

adityasingh2400:
Thanks, both done in `0f9240d`.

The comment in `Atlas.__init__` is gone.

`_ProgressCallback` and `_download_callback` now live in `brainglobe_atlasapi/callback.py`. `core.py` and `bg_atlas.py` both import `_download_callback` from there, and `core.py` no longer needs the `fsspec.callbacks` import at all.

I moved the two tests onto the new module rather than reaching through `core`, since `core` only uses `_download_callback` and keeping `_ProgressCallback` imported there just to satisfy a test would have been an unused import.

The package needs `meshio` and `s3fs` to import, which I do not have here, so I exercised the new module directly instead of running the suite:

```
with handler    -> _ProgressCallback | is _ProgressCallback: True
without handler -> TqdmCallback | is TqdmCallback: True | not _ProgressCallback: True
branched hooks carry fn_update: True
handler received (completed, total): [(42, 100)]
```

`ruff check` passes on the package and both test files. CI will be the real confirmation.

IgorTatarnikov (commented):
I've done a bit of user testing and I think using the file size doesn't work as we're using OME-Zarr therefore `s3fs` downloads many little chunks leading to a confusing progress bar (see below). We should just use the number of files as a proxy.

https://github.com/user-attachments/assets/2bf90162-b936-40e6-bbd7-0e38001c83a0

adityasingh2400:
Good catch on the OME-Zarr chunking, that matches what the video shows. Switched to the file count in `30b4ce4`.

Two things worth flagging from doing it.

`TqdmCallback` replaces `Callback.call` with a version that only drives the bar and never runs `hooks`, so passing a hook to the subclass does nothing at all. That is why the old code had to reach for `branched` with a plain `Callback`. The handler is now invoked from a `call` override instead, which is what makes the top level file count reach `fn_update`:

```python
def call(self, *args, **kwargs):
    super().call(*args, **kwargs)
    self.fn_update(self.value, self.size or 0)
```

The `branched` override is gone, so per-file byte progress is no longer forwarded.

I also expected a two file directory to report a total of 2 and it reports 3, since fsspec counts the directory entry alongside its files. Rather than pin that number the test now asserts the shape, one stable total, a final call of `(total, total)`, and a total that is small and is never the 4096 or 2048 byte sizes the fixture writes.

Verified against real fsspec, running the fetch both ways:

```
recursive=False: calls=[(0, 1), (1, 1)]
recursive=True:  calls=[(0, 3), (1, 3), (2, 3), (3, 3)]
byte transfer within a file reports nothing
```

`ruff check` and `ruff format --check` both pass. I still cannot run the suite locally without `meshio` and `s3fs`, so CI is the real check.

brainglobe_atlasapi/callback.py
```diff
@@ -3,38 +3,36 @@
 from collections.abc import Callable
 from typing import Optional
 
-from fsspec.callbacks import Callback, TqdmCallback
+from fsspec.callbacks import TqdmCallback
 
 
 class _ProgressCallback(TqdmCallback):
```
IgorTatarnikov: With the new structure, I think we can collapse the factory function (`_download_callback`). There could be an if statement inside the overloaded `call` method. This class could then be public and have a public facing name. Perhaps `AtlasCallback`? This would simplify the structure. The docstring for this function can also be cleaned up. It currently contains a lot of detail for how we arrived at this rather than what the class is. This would be confusing to future developers who look at the code. The how process is documented in the PR if anyone wants to refer back! 

```
if self.fn_update:
	sef.fn_update(self.value, self.size or 0)
```

IgorTatarnikov (commented):
Perfect! That's a much better downstream experience. Could you update the docstring for `fn_update` in `BrainGlobeAtlas` to reflect the change in behaviour (from bytes to files)?

adityasingh2400:
Done in `e1b3d8e`.

```
fn_update : Callable
    Handler function to update during download. Takes the number of files
    fetched so far and the total number of files to fetch. An atlas is an
    OME-Zarr store, so a byte count would tick through thousands of small
    chunks and read as a stalling bar.
```

I checked the other `fn_update` docstrings while I was there and deliberately left one alone. `utils.retrieve_over_http` also says "completed and total bytes", but that one is still accurate, it is a plain HTTP download that accumulates `len(chunk)` from `iter_content` rather than going through the fsspec callback. Changing it would have made the docs wrong.

`update_atlases.update_atlas` and `install_atlas` just say "a callback function to update progress during download", which stays correct either way. Happy to make those specific too if you would rather they spell out the units.

brainglobe_atlasapi/callback.py
adityasingh2400: Done in `1b8badd`. `_download_callback` is gone, the class is public as `AtlasCallback`, and `fn_update` now defaults to None with the guard inside `call`.

I trimmed the class docstring to what the class is and moved the one line about `TqdmCallback` not running `hooks` onto `call`, since that is the reason the override exists and it is the thing someone would otherwise undo.

Verified against real fsspec that the call sequences are unchanged:

```
recursive=False: [(0, 1), (1, 1)]
recursive=True:  [(0, 3), (1, 3), (2, 3), (3, 3)]
no handler:      transfers fine, reports nothing
```

brainglobe_atlasapi/bg_atlas.py
```diff
@@ -57,8 +57,10 @@ class BrainGlobeAtlas(core.Atlas):
         this to False to avoid waiting for remote server response on atlas
         instantiation and to suppress warnings.
     fn_update : Callable
-        Handler function to update during download. Takes completed and total
-        bytes.
+        Handler function to update during download. Takes the number of files
+        fetched so far and the total number of files to fetch. An atlas is an
+        OME-Zarr store, so a byte count would tick through thousands of small
+        chunks and read as a stalling bar.
```
IgorTatarnikov:
```suggestion
        Handler function to update during download. Takes the number of files
        fetched so far and the total number of files to fetch.
```

brainglobe_atlasapi/callback.py
```diff
@@ -0,0 +1,41 @@
...
+
+from fsspec.callbacks import TqdmCallback
+
+
+class AtlasCallback(TqdmCallback):
+    """Show a tqdm bar for an atlas download and report progress onwards.
+
+    An atlas is an OME-Zarr store, so this reports the number of files fetched
+    rather than the number of bytes. Bytes would tick through thousands of
+    small chunks and read as a stalling bar.
```
IgorTatarnikov:
```suggestion
    An atlas is an OME-Zarr store, so this reports the number of files fetched
    rather than the number of bytes.
```

brainglobe_atlasapi/core.py
```diff
@@ -817,7 +820,7 @@ def __getitem__(self, key):
                     remote_path,
                     resolution_path,
                     recursive=True,
-                    callback=TqdmCallback(),
+                    callback=AtlasCallback(self.fn_update),
```
IgorTatarnikov, line 829: `self.fn_update` is not defined here. It has to be passed through the constructor from the core `Atlas` class.

tests/atlasapi/test_bg_atlas.py
IgorTatarnikov, line 1: Could you move these tests into a separate submodule (`tests/atlasapi/test_callback.py`)?

tests/atlasapi/test_core_atlas.py
```diff
@@ -573,3 +573,26 @@ def test_get_structure_mask_downloads_once_and_caches(atlas, monkeypatch):
         # Remove the on-demand chunk we materialised; it re-downloads lazily.
         if chunk_dir.exists():
             shutil.rmtree(chunk_dir)
+
+
+def test_core_atlas_shares_the_progress_handler(mocker):
```
IgorTatarnikov: The name of the test doesn't align with what it's testing. No `BrainGlobeAtlas` or `Atlas` objects are instantiated. The test as written doesn't need to be added, We should test that `Atlas` reports downloading the lazy components.

IgorTatarnikov (changes requested):
There's a bug where `AdditionalRefDict` was accessing an attribute that doesn't exist. Could you fix that, and add some tests to cover this path?

I've also made some notes on test organisation in general, I think a lot of the tests should be combined into a separate submodule now that they've been refactored out of `bg_atlas.py` and `core.py`.

brainglobe_atlasapi/core.py
adityasingh2400: Good catch, that was a real bug. `AdditionalRefDict.__getitem__` read `self.fn_update` and the class never set it, so any atlas built with a handler raised `AttributeError` the moment it touched an additional reference.

Fixed in `8dbc6f8` by taking `fn_update` in the constructor and passing it from `Atlas`. The fetch happens long after the owning `Atlas` is built, so the dict has to carry the handler rather than read it off the atlas at download time.

I checked both commits with an AST pass over every `self.fn_update` in `core.py`:

```
before  Atlas: assigned=[72] read=[233,281,342,728]   OK
        AdditionalRefDict: assigned=None read=[823]   read but never assigned
after   Atlas: assigned=[72] read=[150,234,282,343,729]
        AdditionalRefDict: assigned=[765] read=[829]
```

tests/atlasapi/test_bg_atlas.py
adityasingh2400: Moved to `tests/atlasapi/test_callback.py` in `8dbc6f8`, renamed to `test_atlas_callback_*` to match the class. The now unused `fsspec` import is gone from `test_bg_atlas.py`.

tests/atlasapi/test_core_atlas.py
adityasingh2400: Agreed, that test instantiated no atlas and only covered `AtlasCallback`, which `test_callback.py` now owns. Dropped it and added two that do exercise the objects:

`test_atlas_passes_fn_update_to_additional_references` builds a real `core.Atlas` with a handler and asserts it reaches the lazily fetched references, and `test_additional_ref_dict_reports_download_progress` covers the `AdditionalRefDict` path from your other comment.

brainglobe_atlasapi/bg_atlas.py
adityasingh2400: Applied in `8dbc6f8`.

brainglobe_atlasapi/callback.py
adityasingh2400: Applied in `8dbc6f8`.

PR #908, opened 2026-08-05
List locally installed custom atlases

### LINKED ISSUE #906: [Feature] show custom atlases in brainglobe list
at present when we list atlases custom atlases dont show up. it would be nice if they were somewhere in the list, while making it clear they are user installed

adityasingh2400:
Fixes #906.

`get_atlases_lastversions` only kept a downloaded atlas if its name also appeared in `last_versions.conf` or `custom_atlases.conf`. Nothing in this repo writes `custom_atlases.conf`, so an atlas a user installed themselves has files on disk but no conf entry, and is silently dropped from `brainglobe list`.

Every downloaded atlas is now included. User installed ones carry a new `custom` key and get a `Custom` column in the table. A custom atlas with no registered version has nothing to compare against, so its latest version stays empty and `updated` is `None` rather than a red cross claiming it is out of date. The custom conf read moved into a small `get_custom_atlases` helper so the two call sites share it.

Reverting `list_atlases.py` alone gives 6 failed and 12 passed, including a `KeyError` where the custom atlas is missing from the result entirely, and a table assertion showing the old output rendering a red cross in the updated column. After, the full `tests/atlasapi` suite is green at 119 passed, verified at `COLUMNS` of 40, 80 and 200 so the table renders correctly at different widths.

Disclosure: this change was prepared with AI assistance. I have reviewed and tested it.

### DIFF
--- a/brainglobe_atlasapi/list_atlases.py
+++ b/brainglobe_atlasapi/list_atlases.py
@@ -80,6 +80,27 @@ def get_local_atlas_version(atlas_name: str) -> Optional[str]:
         return None
 
 
+def get_custom_atlases() -> Dict[str, Any]:
+    """Read the versions of the locally registered custom atlases.
+
+    Returns
+    -------
+    dict
+        A mapping from custom atlas name to version. Empty if no custom
+        atlases have been registered.
+    """
+    custom_path = (
+        config.get_brainglobe_dir()
+        / "brainglobe-atlasapi"
+        / descriptors.V3_ATLAS_ROOTDIR
+        / "custom_atlases.conf"
+    )
+    try:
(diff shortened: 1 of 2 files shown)


adityasingh2400:
Addressed the SonarCloud issue in 9b4dbd9.

It flagged `"[green]:heavy_check_mark:[/green]"` appearing three times in `list_atlases.py`. Two of those were already there for the downloaded and updated columns, and the custom column this PR adds was the third, so the finding is fairly attributable to this change. Pulled it up into a `CHECK_MARK` constant next to the imports, which also means the three cells cannot drift apart if the markup is ever changed.

black and isort are clean on the file.

PR #925, opened 2026-08-11
set default chunk size for all volumes including masks to 128

PolarBean:
We ran into an issue where packaging was writing large chunk sizes of 128MB. This was causing neuroglancer links to not properly function. Here we write a more sensible chunk size of 128.


brainglobe_atlasapi/atlas_generation/stacks.py
```diff
@@ -90,6 +90,11 @@ def write_multiscale_ome_zarr(
         group=root,
         axes=axes,
         coordinate_transformations=transformations,
+        storage_options={
+            "chunks": (
+                (1, 128, 128, 128) if images[0].ndim == 4 else (128, 128, 128)
```
IgorTatarnikov: The 4D array is where most of the files will be generated. These will never be visualised, which is why I set the chunk size to (1, 256, 256, 256) for that one. If we're differentiating it, perhaps it's worth keeping that in.

IgorTatarnikov (commented):
Just a small suggestion

IgorTatarnikov (approved):
LGTM, thanks!

Issue #852, opened 2026-07-03
[BUG] Tests are failing as NGFF Zarr and OME zarr have been updated


PolarBean:
```python

  tests/atlasgen/conftest.py:30: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  brainglobe_atlasapi/bg_atlas.py:116: in __init__
      super().__init__(self.brainglobe_dir / self.local_full_name)
  brainglobe_atlasapi/core.py:95: in __init__
      self._template_pyramid_level = _determine_pyramid_level(
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  
  multiscale = NgffMultiscales(images=[NgffImage(data=dask.array<from-zarr, shape=(1320, 800, 1140), dtype=uint16, chunksize=(83, 100...rmations=None, omero=None, name='/', type=None, metadata=None, extra={}), scale_factors=None, method=None, chunks=None)
  resolution = (100.0, 100.0, 100.0)
  
      def _determine_pyramid_level(
          multiscale: nz.Multiscales, resolution: Tuple[float, float, float]
      ) -> int:
          for idx, metadata in enumerate(multiscale.metadata.datasets):
              # Only check spatial scale against resolution
  >           scales = metadata.coordinateTransformations[0].scale[-3:]
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  E           AttributeError: 'TransformSequence' object has no attribute 'scale'
  

```

From conversations with @IgorTatarnikov this seems to be due to NGFF zarr updating their API. Our short term solution is to pin OME zarr and NGFF zarr

AmitSubhash:
Is anyone on this? Happy to take it if not.

Would you prefer adapting to the new ngff-zarr metadata layout, or pinning for now? I would confirm which versions broke it first either way.

IgorTatarnikov:
We should aim to support the new versions of `ngff-zarr` and `ome-zarr-py`. Ideally, in a backwards compatible manner, but if not, a lower pin is better than an upper pin!

PR #930, opened 2026-08-14
Unpins ngff-zarr

IgorTatarnikov:
Unpins `ngff-zarr`, `ome-zarr-py` remains pinned.

See #852 

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


pyproject.toml
```diff
@@ -25,8 +25,9 @@ dependencies = [
     "DracoPy",
     "fsspec",
     "meshio",
-    "ngff-zarr<0.37.0",
+    "ngff-zarr>=0.37.0",
     "numpy<2.5",
+    "numpy",
```
adamltyson: Presumably this should be removed?

adamltyson (approved):

pyproject.toml
IgorTatarnikov: Good catch, that was a bad rebase on my part. I'll remove it now.

PR #944, opened 2026-09-01
Add nested metadata structure validation (part 2 of #782)

a-shahrabi:
---

## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**

Second of three PRs for #782. `validate_metadata` only checks the ten flat top-level keys in `METADATA_TEMPLATE`, so nothing currently validates the nested component structure introduced in #769. `validate_atlas_files` reaches into `metadata["template"]["location"]` and friends to build paths, but a missing or malformed component dict there surfaces as a `KeyError` or `TypeError` rather than a clear validation failure.

**What does this PR do?**

Adds `validate_nested_metadata_structure`, which checks that `atlas.metadata` contains `template`, `annotation_set`, `terminology` and `coordinate_space`, that each is a dict with `name`, `version` and `location` string keys, and that the sub-dicts embedded in `annotation_set` (`template`, `terminology`) and `coordinate_space` (`template`) have the same shape. A recursive private helper handles the shared shape, and assertion messages use dotted labels (`annotation_set.template.version`) so failures point at the exact location.

I validated the expected shape against `ComponentInfo.generate_metadata_dict()` in `atlas_packaging_data.py` rather than the `format_*_stub` descriptors, since those three appear to be uncalled (raised on #782).

Two things I left alone but am happy to change:

- `validate_metadata` and this function now both validate metadata, one flat and one nested. I kept them separate so this PR doesn't modify an existing validator, but they could be merged.
- I haven't made `validate_atlas_files` use this, so a malformed component dict there still raises `KeyError` rather than a clean assertion. That felt closer to the PR 3 scope.

## References

Part 2 of 3 for #782. Independent of #879 (part 1), since they touch different functions and can merge in either order.

Six new tests in `tests/atlasgen/test_validation.py`: one asserting the real `kim_dev_mouse_e11-5_mri-adc_31.5um` fixture passes, and five parametrized negative cases covering a missing top-level component, a component that isn't a dict, a missing `location` key, a non-`str` `version`, and a missing nested sub-dict. Each negative case deep-copies the metadata before mutating it and asserts the error message identifies the failing key.

Full suite: 340 passed, 5 xfailed (334 before, plus the 6 new tests). The 5 xfails are unchanged.

## Is this a breaking change?

No. The new validator passes against the existing test atlas.

No.

## Checklist:

- [ ] The documentation has been updated to reflect any changes. Not applicable.
- [x] The code has been formatted with [[pre-commit](https://pre-commit.com/)](https://pre-commit.com/)

---

### DIFF
--- a/brainglobe_atlasapi/atlas_generation/validate_atlases.py
+++ b/brainglobe_atlasapi/atlas_generation/validate_atlases.py
@@ -607,6 +607,114 @@ def validate_metadata(atlas: BrainGlobeAtlas):
     return True
 
 
+def _validate_component_metadata_shape(
+    component: dict, label: str, nested_keys: tuple[str, ...] = ()
+) -> bool:
+    """Validate that a component metadata dict has the expected shape.
+
+    Recursively checks that `component` is a dict containing `name`,
+    `version`, and `location` keys of type `str`, and that any sub-dicts
+    named in `nested_keys` are themselves present and have the same
+    three-key shape.
+
+    Parameters
+    ----------
+    component : dict
+        The component metadata dict to validate (e.g.
+        `atlas.metadata["template"]`).
+    label : str
+        A human-readable label for `component`, used in assertion
+        messages (e.g. "template", "annotation_set.template").
+    nested_keys : tuple of str, optional
+        Names of sub-dicts within `component` that should be recursively
+        validated against the same three-key shape. By default, empty.
+
+    Returns
+    -------
+    bool
+        True if `component` and any nested sub-dicts have the expected
+        shape.
+
+    Raises
+    ------
+    AssertionError
+        If `component` is not a dict, is missing a required key, a key
+        has the wrong type, or a nested sub-dict fails the same checks.
+    """
+    assert isinstance(
+        component, dict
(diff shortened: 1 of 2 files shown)


a-shahrabi:
Added a guard asserting atlas.metadata is a dict, since without it a list or string would pass the key-presence checks via membership/substring matching and then fail with a TypeError rather than a clean assertion. Added tests covering that, using a list and a string that both contain the expected key names.

A few things I deliberately didn't check, since they seem like policy questions rather than bugs. Empty or whitespace-only strings currently pass the str check. Extra unexpected keys in a component are ignored. And the template/terminology embedded in annotation_set aren't checked for consistency with the top-level entries (related to my question on #782). I can add any of these if you'd like them enforced.

Separately, and unrelated to this PR: CI is failing on test_atlas_name_matches_lastversions because nmt_arm_sym_macaque_250um is in last_versions.conf but missing from AtlasName in atlas_name.py. It also fails locally with my changes stashed, and I've rebased onto latest main and it persists, so it'll block any PR until the name is added. Happy to add it here if that's useful, or leave it as a separate fix.

Everything else passes: 341 passed, 5 xfailed.

PR #945, opened 2026-09-01
Add lazy dask atlas loading

### LINKED ISSUE #882: [Feature] Load atlas components as lazy `dask` arrays
Currently, `brainglobe-atlasapi` returns the array components of atlases as in-memory `numpy` arrays. It would be nice to have an optional method for loading the arrays as lazy `dask` objects. Especially now that they are stored as zarr files on disk.

Based on discussion in #141

sdiebolt:
Closes #882.

## Summary
- add `lazy=True` to `BrainGlobeAtlas`/core `Atlas`
- when `lazy=True`, return dask arrays for template, annotation, hemispheres, and structure masks
- make `Atlas`/`BrainGlobeAtlas` generic over array types for proper hints in IDEs and with static type checkers
- keep coordinate lookup methods returning scalar values

### DIFF
--- a/brainglobe_atlasapi/bg_atlas.py
+++ b/brainglobe_atlasapi/bg_atlas.py
@@ -54,8 +54,8 @@ def _version_str_from_tuple(version_tuple: Tuple[int, ...]) -> str:
 
 
 class BrainGlobeAtlas(
-    core.Atlas[core.ReferenceArray, core.AnnotationArray, core.LabelArray],
-    Generic[core.ReferenceArray, core.AnnotationArray, core.LabelArray],
+    core.Atlas[core.TemplateArray, core.AnnotationArray, core.LabelArray],
+    Generic[core.TemplateArray, core.AnnotationArray, core.LabelArray],
 ):
     """Add remote atlas fetching and version comparison functionalities
     to the core Atlas class.
--- a/brainglobe_atlasapi/bg_atlas.py
+++ b/brainglobe_atlasapi/bg_atlas.py
@@ -1,11 +1,24 @@
 """Defines the BrainGlobe Atlas API V3 classes and functions."""
 
+from __future__ import annotations
+
 import re
 from collections.abc import Callable
 from io import StringIO
 from pathlib import Path
-from typing import List, Optional, Tuple, Union
+from typing import (
+    Generic,
+    List,
+    Literal,
+    Optional,
+    Tuple,
+    Union,
+    overload,
+)
 
+import dask.array as da
+import numpy as np
+import numpy.typing as npt
 import s3fs
 from fsspec.callbacks import TqdmCallback
 from rich import print as rprint
@@ -40,7 +53,10 @@ def _version_str_from_tuple(version_tuple: Tuple[int, ...]) -> str:
     return "_".join(str(num) for num in version_tuple)
 
 
-class BrainGlobeAtlas(core.Atlas):
+class BrainGlobeAtlas(
+    core.Atlas[core.ReferenceArray, core.AnnotationArray, core.LabelArray],
+    Generic[core.ReferenceArray, core.AnnotationArray, core.LabelArray],
+):
     """Add remote atlas fetching and version comparison functionalities
     to the core Atlas class.
 
@@ -59,7 +75,54 @@ class BrainGlobeAtlas(core.Atlas):
     fn_update : Callable
         Handler function to update during download. Takes completed and total
         bytes.
+    lazy : bool
+        If True, atlas array properties return dask arrays instead of loading
+        them into memory as numpy arrays.
     """
+
+    @overload
+    def __init__(
+        self: BrainGlobeAtlas[
+            npt.NDArray[np.uint16],
+            npt.NDArray[np.uint32],
+            npt.NDArray[np.uint8],
+        ],
+        atlas_name: AtlasName,
+        version: Optional[str] = None,
+        brainglobe_dir: Optional[Union[str, Path]] = None,
+        check_latest: bool = True,
+        config_dir: Optional[Union[str, Path]] = None,
+        fn_update: Optional[Callable] = None,
+        lazy: Literal[False] = False,
+    ) -> None: ...
+
+    @overload
+    def __init__(
+        self: BrainGlobeAtlas[da.Array, da.Array, da.Array],
+        atlas_name: AtlasName,
+        version: Optional[str] = None,
+        brainglobe_dir: Optional[Union[str, Path]] = None,
+        check_latest: bool = True,
+        config_dir: Optional[Union[str, Path]] = None,
+        fn_update: Optional[Callable] = None,
+        lazy: Literal[True] = True,
+    ) -> None: ...
+
+    @overload
+    def __init__(
+        self: BrainGlobeAtlas[
+            Union[npt.NDArray[np.uint16], da.Array],
+            Union[npt.NDArray[np.uint32], da.Array],
+            Union[npt.NDArray[np.uint8], da.Array],
+        ],
+        atlas_name: AtlasName,
+        version: Optional[str] = None,
+        brainglobe_dir: Optional[Union[str, Path]] = None,
+        check_latest: bool = True,
+        config_dir: Optional[Union[str, Path]] = None,
+        fn_update: Optional[Callable] = None,
+        lazy: bool = False,
+    ) -> None: ...
 
     def __init__(
         self,
@@ -69,6 +132,7 @@ class BrainGlobeAtlas(core.Atlas):
         check_latest: bool = True,
         config_dir: Optional[Union[str, Path]] = None,
         fn_update: Optional[Callable] = None,
+        lazy: bool = False,
     ):
         self._remote_version = None
         self._local_full_name = None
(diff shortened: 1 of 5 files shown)


PolarBean:
This seems to be quite a lot of logic to ensure the type checker understands that the typing of template and annotation are correlated. I understand this is strictly more correct but I don't know practically what we gain from it. 

I think the simpler union typing here would be better but I would also like to hear @IgorTatarnikov's opinion

sdiebolt:
> This seems to be quite a lot of logic to ensure the type checker understands that the typing of template and annotation are correlated. I understand this is strictly more correct but I don't know practically what we gain from it.

This avoids typing errors; if we use a union, any call to e.g. `.compute()` will lead to a typing error since type checkers cannot narrow down to a dask array. This also means IDEs won't properly auto-complete on `annotation`/`template`/`hemisphere`, and users/developers using the package will need to narrow the type manually using `cast` to please type checkers.

But I totally agree generics are somewhat verbose... So very happy to follow your advice.

brainglobe_atlasapi/core.py
```diff
@@ -399,6 +464,8 @@ def hemisphere_from_coords(
 
         """
         hem = self.hemispheres[self._idx_from_coords(coords, microns)]
+        if self.lazy and hasattr(hem, "compute"):
```
PolarBean: when would hemisphere not have compute if lazy is False?

brainglobe_atlasapi/core.py
```diff
@@ -432,6 +499,8 @@ def structure_from_coords(
             Structure containing the coordinates.
         """
         rid = self.annotation[self._idx_from_coords(coords, microns)]
+        if self.lazy and hasattr(rid, "compute"):
```
PolarBean: same comment as above

tests/atlasapi/test_core_atlas_lazy.py
```diff
@@ -0,0 +1,88 @@
...
+    atlas = _atlas(tmp_path, monkeypatch, data)
+
+    assert isinstance(atlas.template, da.Array)
+    assert isinstance(atlas.annotation, da.Array)
+    assert isinstance(atlas.hemispheres, da.Array)
+    assert atlas.template.compute()[0, 0, 1] == 1
+    assert atlas.annotation.compute()[0, 1, 0] == 2
+    assert atlas.hemispheres.compute()[0, 0, 1] == atlas.left_hemisphere_value
+
+
```
PolarBean, line 73: None of the tests actually test using the example atlas with lazy=True and ensuring it behaves as expected

tests/atlasapi/test_core_atlas_lazy.py
```diff
@@ -0,0 +1,88 @@
...
+    assert structure == 2
+    assert atlas.structure_from_coords((0, 0, 0), as_acronym=True) == "leaf"
+    assert atlas.hemisphere_from_coords((0, 0, 1), as_string=True) == "left"
+
+
+def test_lazy_get_structure_mask_returns_dask_array(tmp_path, monkeypatch):
+    """lazy=True returns a dask array for structure masks."""
+    data = da.from_array(np.ones((1, 2, 2, 2), dtype=np.uint8))
+    atlas = _atlas(tmp_path, monkeypatch, data)
+
```
PolarBean, line 91: we should add a small comment in either the docs or the README about using the API with lazy=True. What this returns and how the user is expected to call .compute()

PolarBean (changes requested):

brainglobe_atlasapi/core.py
sdiebolt: Indeed, that was from a first test where my type checker was complaining, but actually not useful anymore!

tests/atlasapi/test_core_atlas_lazy.py
sdiebolt: Added `test_example_atlas_lazy_properties_are_dask_arrays`, which instantiates `BrainGlobeAtlas(..., lazy=True)` using the example atlas and checks `template`/`annotation`/`hemispheres` return dask arrays and compute expected values.

tests/atlasapi/test_core_atlas_lazy.py
sdiebolt: I added a short README note. Maybe we should also add it to the docs? If so, where would it be most appropriate?

brainglobe_atlasapi/core.py
```diff
@@ -38,6 +44,16 @@
     read_json,
 )
 
+ReferenceArray = TypeVar(
```
alessandrofelder: We are moving towards consistently using "template" rather than "reference" so I think this should be "TemplateArray".

tests/atlasapi/test_core_atlas_lazy.py
alessandrofelder: Maybe a new section here: https://brainglobe.info/api_examples/brainglobe_atlasapi.html ?

alessandrofelder:
> But I totally agree generics are somewhat verbose... So very happy to follow your advice.

Tricky! Following some discussion with @IgorTatarnikov, my two-cents would be: (While I would be against introducing this level of complexity in more user-facing code e.g. packaging scripts or `brainrender` example script,) in this case I am leaning towards having more precise typehints for a critical class that hopefully even power-users rarely need to look inside - to aid our typecheckers - accepting the complexity that comes with it.

(As Igor points out, this might also be an opportunity for the core team to learn more about type hints as Python seems in general to be becoming more strongly typed)

Also added some smaller review comments.

I've opened #946 to discuss reducing complexity in this part of the codebase in a different way, but that's a separate PR.

brainglobe_atlasapi/core.py
sdiebolt: Indeed, still had the old "reference" term in mind, I'll fix that.

tests/atlasapi/test_core_atlas_lazy.py
sdiebolt: These notebooks are hosted on another repo, right? Should I open a PR there?

sdiebolt:
Thanks @PolarBean and @alessandrofelder for your reviews! I tried to address your comments. About documenting this feature: should I open a new PR in the docs repository?

tests/atlasapi/test_core_atlas_lazy.py
alessandrofelder: Yes, please :)

brainglobe_atlasapi/bg_atlas.py
```diff
@@ -59,8 +75,55 @@ class BrainGlobeAtlas(core.Atlas):
...
+        If True, atlas array properties return dask arrays instead of loading
+        them into memory as numpy arrays.
     """
 
+    @overload
+    def __init__(
+        self: BrainGlobeAtlas[
+            npt.NDArray[np.uint16],
+            npt.NDArray[np.uint32],
+            npt.NDArray[np.uint8],
```
IgorTatarnikov, line 88: I think these should reuse the constants defined in `descriptors.py`. Much easier to change in the future if we ever decided to be more flexible in the allowed dtypes.

brainglobe_atlasapi/core.py
```diff
@@ -38,6 +44,16 @@
     read_json,
 )
 
+TemplateArray = TypeVar(
+    "TemplateArray", bound=Union[npt.NDArray[np.uint16], da.Array]
+)
+AnnotationArray = TypeVar(
+    "AnnotationArray", bound=Union[npt.NDArray[np.uint32], da.Array]
+)
+LabelArray = TypeVar(
```
IgorTatarnikov, line 53:
```suggestion
MaskArray = TypeVar(
```

Small variable name suggestion, `LabelArray` is really close to `AnnotationArray` whereas `MaskArray` gives a bit more semantic context. It might be a bit of a stretch for the hemispheres array as it uses 0, 1,  and 2. I'm not too fussed either way. Open to discussion @PolarBean, @alessandrofelder.

brainglobe_atlasapi/core.py
```diff
@@ -82,19 +98,52 @@ def _determine_pyramid_level(
...
     left_hemisphere_value = 1
     right_hemisphere_value = 2
 
-    def __init__(self, path):
+    @overload
+    def __init__(
+        self: Atlas[
+            npt.NDArray[np.uint16],
+            npt.NDArray[np.uint32],
+            npt.NDArray[np.uint8],
```
IgorTatarnikov, line 121: As in `bg_atlas.py` this should use the constants from `descriptors.py`.

IgorTatarnikov (commented):
Looks good to me! Just a few comments from my side.

A proposal in terms of balancing complexity and specificity. Perhaps, we can drop the specific dtype for static typing and rely on `np.integer` instead. This would allow a generic `AtlasArray` that's either `npt.NDArray[np.integer]` or `da.Array`. At runtime we can still enforce the data is cast to the correct type using the `descriptors.*_DTYPE` callers (I know this is done in packaging, I'm not sure if there's a live check during atlas instantiation).

PolarBean:
We also had a chat in the BrainGlobe dev meeting and feel like lazy isnt the best argument name as it could refer to
Dask or retrieving only the parts of components that are needed for an analysis (ie indexing a plane and retrieving only that plane). A clearer name could be dask=True . But open to suggestions!

PR #955, opened 2026-09-07
ci: standardize pre-commit config

AlgoFoe:
Standardizes the pre-commit configuration across BrainGlobe repositories while preserving repo-specific hooks and settings.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,3 +24,28 @@ repos:
       - id: ruff-format
         args:
           - --config=pyproject.toml
+
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.13.0
+    hooks:
+      - id: mypy
+        additional_dependencies:
+          - types-setuptools
+
+  - repo: https://github.com/mgedmin/check-manifest
+    rev: "0.50"
+    hooks:
+      - id: check-manifest
+        args: [--no-build-isolation]
+        additional_dependencies:
+          - setuptools>=77
+          - wheel
+          - setuptools-scm[toml]>=8
+
+  - repo: https://github.com/codespell-project/codespell
+    # Configuration for codespell is in pyproject.toml
+    rev: v2.4.3
+    hooks:
+      - id: codespell
+        additional_dependencies:
+          - tomli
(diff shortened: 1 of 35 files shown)


alessandrofelder:
@brainglobe/active-devs looks like standardising the pre-commit config has the potential to come with lots of (`ruff-format`-induced) formatting edits. I think this is fine, but feel free to disagree.

Separately, do you think there is a case for excluding
* `atlas_scripts/`
* the example notebook

from the formatter?

adamltyson:
> @brainglobe/active-devs looks like standardising the pre-commit config has the potential to come with lots of (ruff-format-induced) formatting edits. I think this is fine, but feel free to disagree.

I think this is fine, and sort of inevitable if we want to standardise linting. 

>Separately, do you think there is a case for excluding
> atlas_scripts/
> the example notebook

I think we should reformat the atlas_scripts. they're code just the same, and long-term not excluding them is marginally simpler. I would also format the notebook, assuming the formatting doesn't do anything weird (as IIRC black used to)

IgorTatarnikov (commented):
We should remove the `tool.black` section in `pyproject.toml` now that it's not being used anymore.

I also noticed that the [cookiecutter](https://github.com/neuroinformatics-unit/python-cookiecutter) has mypy, check-manifest, and codespell. Is this something we also want to introduce here?

```
    - repo: https://github.com/pre-commit/mirrors-mypy
      rev: v1.13.0
      hooks:
          - id: mypy
            additional_dependencies:
                - types-setuptools
    - repo: https://github.com/mgedmin/check-manifest
      rev: "0.50"
      hooks:
          - id: check-manifest
            args: [--no-build-isolation]
            additional_dependencies:
              - setuptools>=77
              - wheel
              - setuptools-scm[toml]>=8
    - repo: https://github.com/codespell-project/codespell
      # Configuration for codespell is in pyproject.toml
      rev: v2.4.3
      hooks:
      - id: codespell
        additional_dependencies:
        - tomli 
```

adamltyson:
>  Is this something we also want to introduce here?

I would say yes, but mypy and codespell may open up a can of worms of other work to do to make them happy!

AlgoFoe:
Yes. I'll update this PR accordingly, and also update the cross-repo rollout [script ](https://github.com/neuroinformatics-unit/scripts/pull/21) to include these hooks along with `pyproject.toml`.

